PR #243, opened 2021-11-10
3D nearest algorithm and parcel merging

matt-frey:


src/3d/parcels/parcel_nearest.f90
```diff
@@ -138,41 +140,47 @@ subroutine find_nearest(isma, iclo, nmerge)
...
+                            ijk = 1 + mod(nx + ix, nx) + nx * mod(ny + iy, ny) + nx * ny * iz
+                            ! Search small parcels for closest other:
+                            do k = kc1(ijk), kc2(ijk)
+                                n = node(k)
+                                if (n .ne. is) then
+                                    delz = parcels%position(n, 3) - z_small
+                                    if (delz*delz < dsqmin) then
+                                        ! works across periodic edge
+                                        delx = get_delx(parcels%position(n, 1), x_small)
+                                        dely = get_dely(parcels%position(n, 2), y_small)
```
sjboeing, line 169: It is possible to do another check on e.g. delx*delx+delz+delz, not sure if it matters much since the coordinates sit close in memory.

matt-frey: I am not sure if we should add another if-statement.

PR #248, opened 2021-11-15
3D vorticity tendency

matt-frey:


src/3d/epic3d.f90
```diff
@@ -79,15 +80,17 @@ subroutine pre_run
...
 
             call init_parcels(field_file, field_tol)
 
             call ls_rk4_alloc(max_num_parcels)
 
-            call init_inversion
+            call init_fft
 
-            call init_parcel_correction
+!             call init_parcel_correction
```
daviddritschel, line 93: Does this not need initialization any more?

src/3d/inversion/inversion.f90
```diff
@@ -199,10 +207,114 @@ subroutine vor2vel(vortg,  velog,  svelog)
...
+
+            call diffx(svelog(0:nz, :, :, 3), ds)   ! w_x = dw/dx in spectral space
+            call fftxys2p(ds, velgradg(:, :, :, 7)) ! w_x in physical space
+
+            ! du/dz = \omegay + dw/dx
+            velgradg(:, :, :, 3) = vortg(:, :, :, 2) + velgradg(:, :, :, 7)
+
+            ! y & z components:
+            call diffx(svelog(0:nz, :, :, 2), ds)   ! v_x = dv/dx in spectral space
+            call fftxys2p(ds, velgradg(:, :, :, 4)) ! v_x in physical space
```
daviddritschel, line 242: Could use dv/dx = \omegaz - du/dy if this is faster here

src/3d/inversion/inversion.f90
```diff
@@ -199,10 +207,114 @@ subroutine vor2vel(vortg,  velog,  svelog)
...
+            ! Compute spectral buoyancy (bs):
+            call fftxyp2s(b, bs)
+
+            call diffy(bs, ds)                      ! b_y = db/dy in spectral space
+            call fftxys2p(ds, db)                   ! db = b_y in physical space
+
+            !$omp parallel
+            !$omp workshare
+            vtend(:, :, :, 1) =  vortg(:, :, :, 1)           * velgradg(:, :, :, 1) & ! \omegax * du/dx
+                              + (vortg(:, :, :, 2) + ft_cor) * velgradg(:, :, :, 2) & ! \omegay * du/dy
```
daviddritschel, line 285: Won’t it be faster to have a code block without f_cor etc when there is no rotation?  Won’t the present code be unnecessarily costly when f … are zero?

src/3d/utils/options.f90
```diff
@@ -42,6 +42,18 @@ module options
...
 
+    !
+    ! physical input
+    !
+
+    !FIXME parameters for coriolis and mean wind
+    !non-dimensional ang_vel of earth = t_scale*Omega
+    !t_scale = 142.8571428571 Omega = 7.2921159e-5
+    logical          :: l_coriolis  = .false.
+    double precision :: ang_vel     = 0.0104173d0       ![m/s] angular velocity
```
daviddritschel, line 53: This should be 2 * pi / 86400

daviddritschel (commented):
Just a few comments.  Looking good!

src/3d/epic3d.f90
matt-frey: No, the FFT is initialized globally. We can remove this.

src/3d/inversion/inversion.f90
matt-frey: True.

src/3d/inversion/inversion.f90
matt-frey: I am not sure ..- I think the compiler will "inline" all of this with optimization `O3`.

src/3d/utils/options.f90
matt-frey: Okay .. I have the values from PMPIC.

src/3d/inversion/inversion.f90
matt-frey: I tested in a separate program ... there is no performance penalty.

src/3d/inversion/inversion.f90
```diff
@@ -1,31 +1,39 @@
 module inversion_mod
     use inversion_utils
     use parameters, only : nx, ny, nz
-    use constants, only : zero, three, four
+    use phys_parameters, only : ft_cor, f_cor
```
sjboeing, line 4: There may be other parameters (like the strength of gravity, and later on the latent heat of condensation) that would be best put in the same module.

matt-frey: Sure. I want to add once we need them. I do not want to add parameters that are not needed yet.

matt-frey: We also have a `phys_constants.f90` file.

matt-frey: https://github.com/matt-frey/epic/blob/3d-vorticity-tendency/src/utils/phys_constants.f90

matt-frey: Some may move to `phys_parameters.f90` though.

sjboeing: Ah yes, it makes sense to make the angular velocity of the earth a constant, but the latitude a parameter.

src/3d/inversion/inversion.f90
```diff
@@ -199,10 +207,135 @@ subroutine vor2vel(vortg,  velog,  svelog)
...
+            call diffy(svelog(0:nz, :, :, 1), ds)      ! u_y = du/dy in spectral space
+            call fftxys2p(ds, velgradg(0:nz, :, :, 2)) ! u_y in physical space
+
+            call diffx(svelog(0:nz, :, :, 3), ds)      ! w_x = dw/dx in spectral space
+            call fftxys2p(ds, velgradg(0:nz, :, :, 7)) ! w_x in physical space
+
+            ! use symmetry to fill z grid points outside domain:
+            ! u_x(-1) =  u_x(1) and u_x(nz+1) =  u_x(nz-1)
+            ! u_y(-1) =  u_y(1) and u_y(nz+1) =  u_y(nz-1)
+            ! w_x(-1) = -w_x(1) and w_x(nz+1) = -w_x(nz-1)
```
sjboeing, line 240: Please check the sign.

src/3d/inversion/inversion.f90
```diff
@@ -199,10 +207,135 @@ subroutine vor2vel(vortg,  velog,  svelog)
...
+
+            call diffy(svelog(0:nz, :, :, 2), ds)      ! v_y = dv/dy in spectral space
+            call fftxys2p(ds, velgradg(0:nz, :, :, 5)) ! v_y in physical space
+
+            call diffy(svelog(0:nz, :, :, 3), ds)      ! w_y = dw/dy in spectral space
+            call fftxys2p(ds, velgradg(0:nz, :, :, 8)) ! w_y in physical space
+
+            ! use symmetry to fill z grid points outside domain:
+            ! v_y(-1) =  v_y(1) and v_y(nz+1) =  v_y(nz-1)
+            ! w_y(-1) = -w_y(1) and w_y(nz+1) = -w_y(nz-1)
```
sjboeing, line 265: Please check the sign.

sjboeing (commented):
I was wondering about two minus signs. Otherwise this looks good (I haven't checked the tests in detail, but they make sense).

src/3d/inversion/inversion.f90
sjboeing: (it doesn't look consistent with the code below)

matt-frey: Yes, I forgot to add the minus sign in the code!

matt-frey: Updated.

src/3d/inversion/inversion.f90
matt-frey: I added the missing minus signs.

sjboeing: Thanks Matthias, I think with that these requests are ready.

PR #249, opened 2021-11-17
3D vorticity tendency less memory

matt-frey:
This pull request reduces the amount of memory used for an EPIC simulation. Instead of 9 elements of the velocity gradient tensor per grid point, only 5 elements are stored. It makes use of the assumption of incompressibility and the relation between vorticity and spatial velocity derivatives.

### DIFF
--- a/src/3d/fields/fields.f90
+++ b/src/3d/fields/fields.f90
@@ -19,9 +19,16 @@ module fields
         vortg,     &   ! vorticity vector field
         vtend,     &   ! vorticity tendency
         velgradg       ! velocity gradient tensor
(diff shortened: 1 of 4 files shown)


matt-frey:
@daviddritschel @sjboeing I let you decide if we want to keep all 9 elements of the velocity gradient tensor per grid point or if we only keep the necessary 5 elements and compute the rest on the fly. I am fine with both solutions.

sjboeing (commented):
This looks right to me. As I said earlier, I think it may be beneficial to give the indices names to clarify what is going on.

matt-frey:
@sjboeing I opened an issue for this (see #251).

PR #266, opened 2021-12-14
Optimizations

matt-frey:
This pull request optimizes
* the tri-linear interpolation
* the Jacobi solver
* grid2par to centre point only

### DIFF
--- a/src/utils/jacobi.f90
+++ b/src/utils/jacobi.f90
@@ -11,7 +11,7 @@ module jacobi
         integer, parameter :: n = 3 ! we have 3x3 matrices
         double precision, parameter :: atol = 1.0e-15
 
-    public :: jacobi_diagonalise
+    public :: jacobi_diagonalise, jacobi_eigenvalues
 
     contains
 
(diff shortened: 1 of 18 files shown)


src/utils/jacobi.f90
```diff
@@ -122,99 +132,213 @@ end subroutine apply_rotation
...
+                    endif
                 enddo
-            endif
-
+            enddo
 
             ! sum of off-diagonal entries
             ! sm should convergence to zero
-            sm = dabs(A(1, 2)) + dabs(A(1, 3)) + dabs(A(2, 3))
+            sm = dabs(A(2, 1)) + dabs(A(3, 1)) + dabs(A(3, 2))
```
sjboeing, line 154: Just wondering if the change in indices here is an optimisation.

matt-frey: No, not really. I wanted to consider only the upper triangular matrix, but it is not yet working.

matt-frey: I need to look into this again more carefully.

matt-frey: Do you agree with this PR otherwise?

sjboeing: Yes, happy with the pull request, but it would indeed be good to figure out what is going on with the off-diagonal entries.

sjboeing: Was just compiling and running it locally

sjboeing: Happy for you to pull it through. I think we are getting there in terms of computational cost (I estimate we are now a factor 6 or so more expensive than PMPIC in terms of cost per parcel per RK4 substep. The par2grid and vol2grid routines would still be the main targets for future optimisation).

matt-frey: This sounds promising. Yes, I agree on par2grid and vol2grid. I will look into this again later. I want to set up the moisture example and then come back again.

PR #267, opened 2021-12-15
3D moisture plume test case

matt-frey:


src/utils/phys_constants.f90
```diff
@@ -9,10 +9,10 @@ module phys_constants
     double precision, parameter :: gravity = 9.81d0
 
     ![J/kg] latent heat of condensation
-    double precision, parameter :: L_c = 2.5e6
+    double precision, parameter :: L_c = 2.501e6
```
sjboeing, line 12: L_v is the more common naming (for vaporisation)

models/3d/moist_3d.f90
```diff
@@ -0,0 +1,147 @@
...
+            allocate(humg(0:nz, 0:ny-1, 0:nx-1))
+
+            buoyg = zero
+            humg = zero
+
+            do i = 0, nx-1
+                do j = 0, ny-1
+                    do k = 0, nz
+
+                        pos = origin + dx * dble((/i, j, k/))
```
sjboeing, line 113: Just a note that the origin needs to be at the surface for this case to work

sjboeing: As well as in the centre of the domain (we can generalise this later)

src/utils/phys_constants.f90
matt-frey: Ok!

models/3d/moist_3d.f90
matt-frey: Ok.

sjboeing (approved):
Looks good to me. I will try to run it and check if the output looks plausible

PR #276, opened 2021-12-21
Store tau and eta

matt-frey:


sjboeing:
Hi @matt-frey: I was wondering about merging in this branch, the changes to detect parcel points that add up in the same grid cell, and some latest changes. Is this one ready to merge?

matt-frey:
@sjboeing I think so. Do we want to apply this change now? If yes, it makes sense to merge this branch first and then do a new pull request.

matt-frey:
@sjboeing I added another option (cannot be set by the user though) in order to catch the restart case with tau and eta. See also https://github.com/matt-frey/epic/pull/276/commits/624ae1e490a875aa9529810554884b5a34d5668d. I think this PR is ready for review.

sjboeing (approved):
Happy to merge this in. Thanks for also doing the initialisation.

PR #279, opened 2022-01-13
Fix calculation of get_delx and get_dely for non-centred domains

matt-frey:
Closes #277.

PR #283, opened 2022-01-17
Fix get_delx for non-centred domains

matt-frey:
@sjboeing I fix this now also in the main branch. It is already fixed in the develop branch (see #279), but this is not yet merged with the main branch.

### DIFF
--- a/src/parcels/parcel_container.f90
+++ b/src/parcels/parcel_container.f90
@@ -39,7 +39,7 @@ module parcel_container
 
             delx = x1 - x2
             ! works across periodic edge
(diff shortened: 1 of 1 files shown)


sjboeing:
Thanks, yes happy for you to merge this. I remember you made another change for non-centered domain. Is that also needed here?

matt-frey:
I only changed this routine in 2D, but it needs to be changed in 3D too (see https://github.com/matt-frey/epic/pull/279/files). This is not needed here.

sjboeing (approved):

PR #284, opened 2022-01-17
Improvement of kinetic energy accuracy

matt-frey:


sjboeing:
I am wondering if, since the KE can be hard to tell apart from the one calculated directly, we should stick to this approach or revert  to the simpler one. Have we seen test cases where it significantly changes the energy budget?

matt-frey:
@sjboeing @daviddritschel According to our discussion I close this pull request.

PR #286, opened 2022-01-17
Index based

sjboeing:
Hi @matt-frey. Here is the work that stores the contributions to the grid in case support points correspond to the same index. Let me know if you can check it for the issue you spotted before.

### DIFF
--- a/src/3d/parcels/parcel_interpl.f90
+++ b/src/3d/parcels/parcel_interpl.f90
@@ -383,7 +383,7 @@ module parcel_interpl
             jj(1) = floor(xyz(2))
             kk(1) = floor(xyz(3))
 
-            call trilinear_weights_inner(xyz, ii(1), jj(1), kk(1), ww)
+            call get_weights(xyz, ii(1), jj(1), kk(1), ww)
 
             ii(1) = mod(ii(1) + nx, nx)
             jj(1) = mod(jj(1) + ny, ny)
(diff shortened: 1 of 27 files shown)


sjboeing:
My changes to the moist setup are based on this one, so would be best to merge this one first if all is OK.

sjboeing:
One thing that originally did not work was OpenMP (I had to fix line length).

matt-frey:
I ran with OpenMP. Yes, I remember a line length issue.

matt-frey:
@sjboeing Have you tested that you still get the same results?

src/3d/parcels/parcel_interpl.f90
```diff
@@ -326,74 +375,98 @@ pure subroutine trilinear(pos, ii, jj, kk, ww)
...
+            double precision, intent(inout)  :: ww(ngp)
+            integer, intent(in)              :: i_lower, j_lower, k_lower
+            double precision                 :: xyz(3)
+            xyz = (pos - lower) * dxi
+            call trilinear_weights_inner(xyz, i_lower, j_lower, k_lower, ww)
+
+        end subroutine trilinear_weights_add
+
+
+        pure subroutine trilinear_weights_inner(xyz, i_lower, j_lower, k_lower, ww)
```
matt-frey, line 439: Can we maybe rename this to `get_weights` or similar? I would also change `i_lower` etc. to just `i`.

src/3d/parcels/parcel_interpl.f90
```diff
@@ -326,74 +375,98 @@ pure subroutine trilinear(pos, ii, jj, kk, ww)
...
+
+            pz = xyz(3) - dble(k_lower)
+            pzc = one - pz
+
+            w00 = pyc * pxc
+            w10 = pyc * px
+            w01 = py * pxc
+            w11 = py * px
+
+            ww(1) = ww(1)+pzc * w00
```
matt-frey, line 461: Please add spaces before and after `+`. Thanks.

matt-frey (changes requested):
I have two remarks.

sjboeing:
Just revised the subroutines. Let me know if it is ready to merge.

src/3d/parcels/parcel_interpl.f90
sjboeing: Changed it to get_weights, which is indeed more explicit, with indices i, j, k (also used this naming in trilinear_weights_add now).

src/3d/parcels/parcel_interpl.f90
sjboeing: Done this.

matt-frey: Thanks!

matt-frey (approved):
@sjboeing I fixed a few more white spaces (with https://github.com/matt-frey/epic/pull/286/commits/6f8c49e8c189d6423ca6606eda181a7e65115e07). Thanks a lot for working on this!

PR #287, opened 2022-01-18
Use proper interpolation name

matt-frey:
We currently use the name `trilinear` in 2D, but it should be named `bilinear` instead.


sjboeing (approved):
Thanks, @matt-frey!

PR #288, opened 2022-01-19
Moist test debug pull request in development

sjboeing:
Hi @matt-frey: here is a draft request for the changes for the moist run. I will add some comments from my side in a bit.


configure.ac
```diff
@@ -239,7 +239,7 @@ else
     AC_MSG_RESULT([no])
 fi
 
-ENABLE_DRY_MODE='yes'
+ENABLE_DRY_MODE='no'
```
sjboeing, line 242: This should be default, I think.

convert_fields.py
```diff
@@ -0,0 +1,63 @@
+# transpose the fields too!
```
sjboeing, line 1: Simple script to convert hdf5 fields to netcdf.

examples/moist.nml
```diff
@@ -7,7 +7,7 @@
  box%ncells = 32, 32, 32
 
  box%extent = 6280, 6280, 6280
- box%origin = 0, 0, 0
+ box%origin = -3140, -3140, 0
```
sjboeing, line 10: Centering this configuration for now. We should also test non-centred domains at some point though.

src/3d/epic3d.f90
```diff
@@ -22,9 +22,9 @@ program epic3d
     use inversion_mod, only : vor2vel_timer, vtend_timer
     use inversion_utils, only : init_fft
     use parcel_interpl, only : grid2par_timer, par2grid_timer
-#ifndef NDEBUG
-    use parcel_interpl, only : sym_vol2grid_timer
-#endif
+!#ifndef NDEBUG
```
sjboeing, line 25: Could also remove this altogether in 3D.

configure.ac
matt-frey: We had dry mode as default because of the 2D paper, but we can change now of course. We can also remove this flag later.

src/3d/epic3d.f90
matt-frey: Yes, it makes sense to remove this for the 3D version.

examples/moist.nml
matt-frey: We should definitelty test the non-centered version as well, and only merge if it works as well.

sjboeing: I would merge this first, so it works and we can run the simulations. We can then test/work on a non-centred version in a separate PR/issue.

third_order_conversion.py
```diff
@@ -0,0 +1,332 @@
+from numba import njit, prange, set_num_threads
```
sjboeing, line 1: I would like to do some more work on this, but have prioritised having a working version for now.

examples/moist.nml
sjboeing: Just tested again. For now, it only works centred. I have a debugging branch somewhere which I will update.

matt-frey: Okay. I am wondering why it still not working for non-centred domains. :thinking:

matt-frey (approved):

PR #289, opened 2022-01-19
Periodicity fixes

sjboeing:
Here is a suggested fix

### DIFF
--- a/examples/moist.nml
+++ b/examples/moist.nml
@@ -7,7 +7,7 @@
  box%ncells = 32, 32, 32
 
  box%extent = 6280, 6280, 6280
- box%origin = -3140, -3140, 0
+ box%origin = 0, 0, 0
 
  moist%H          = 0.8
  moist%z_c        = 2500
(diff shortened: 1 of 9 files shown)


examples/moist.nml
```diff
@@ -7,7 +7,7 @@
  box%ncells = 32, 32, 32
 
  box%extent = 6280, 6280, 6280
- box%origin = -3140, -3140, 0
+ box%origin = 0, 0, 0
```
sjboeing, line 10: Using this now, so it is easy to test the periodicity.

src/2d/parcels/parcel_bc.f90
```diff
@@ -16,7 +16,7 @@ module parcel_bc
         ! @param[inout] position vector of parcel
         subroutine apply_periodic_bc(position)
             double precision, intent(inout) :: position(2)
-            position(1) = position(1) - extent(1) * dble(int((position(1) - center(1)) * hli(1)))
+            position(1) = position(1) - extent(1) * dble(floor((position(1) - lower(1)) * extenti(1)))
```
sjboeing, line 19: Changes to parcel_bc are not strictly needed (they should be the same unless a parcel is far outside the boundaries), but do improve consistency and correctness, I think.

PR #291, opened 2022-01-20
New structure in main like in branch develop

matt-frey:
This PR cherry-picks following commits from the branch develop:
* https://github.com/matt-frey/epic/commit/05c38f90562021e13a6d40bd823fcaace547c631
* https://github.com/matt-frey/epic/commit/d1c586aae3d347f5af9aeee4c8debe8220c36d0f
* https://github.com/matt-frey/epic/commit/4704336143d9a09269e6d3d5ed5d81801cefe0a0
* https://github.com/matt-frey/epic/commit/3508610db975651d7658294e2e9f5d8f4854b9ab

Some other changes had to be done as well.


matt-frey:
@sjboeing I updated the 2d code of the main branch. It has now the latest code from the develop branch and the same directory structure.

src/2d/epic2d.f90
```diff
@@ -124,9 +137,11 @@ subroutine run
 #ifdef ENABLE_VERBOSE
             use options, only : verbose
 #endif
-            double precision :: t    = zero ! current time
+            double precision :: tinit       ! initial time
             integer          :: cor_iter    ! iterator for parcel correction
 
+            tinit = t
```
daviddritschel, line 143: Don't you want t = tinit here?

daviddritschel (commented):
Just one possible error in a quick check

src/2d/epic2d.f90
matt-frey: No, `t` is initially zero, but in a restart simulation `t` is set here https://github.com/matt-frey/epic/blob/develop/src/2d/epic2d.f90#L100. In the 3D version I added `time%initial`, https://github.com/matt-frey/epic/blob/develop/src/3d/epic3d.f90#L94. I actually like the 3D approach better. I can change to this approach in the 2D version too.

matt-frey: I will change in another PR.

matt-frey: Changed in `main` and `develop` with #295 and #294.

PR #292, opened 2022-01-20
Work in progress. Ensure merger position inside domain

sjboeing:
Here is the branch that fixes the merger position. It runs, but I don't seem to get the exact same results as with my other fix, need to check up on this.

### DIFF
--- a/src/3d/parcels/parcel_container.f90
+++ b/src/3d/parcels/parcel_container.f90
@@ -4,7 +4,7 @@
 ! =============================================================================
 module parcel_container
     use options, only : verbose
-    use parameters, only : extent, hli, center
+    use parameters, only : extent, hli, center, upper, lower
     implicit none
 
     integer :: n_parcels
@@ -40,6 +40,11 @@ module parcel_container
             double precision             :: delx
 
             delx = x1 - x2
+#ifndef NDEBUG
+            if((x1 < lower(1) .or. (x2 < lower(1)) .or. (x1 > upper(1)) .or. (x2 > upper(1)) ) then
+                write(*,*) 'point outside domain was fed into get_dely'
+            endif
+#endif
             ! works across periodic edge
             delx = delx - extent(1) * dble(int(delx * hli(1)))
         end function get_delx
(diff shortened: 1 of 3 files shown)


src/3d/parcels/parcel_container.f90
```diff
@@ -40,6 +40,11 @@ elemental function get_delx(x1, x2) result (delx)
             double precision             :: delx
 
             delx = x1 - x2
+#ifndef NDEBUG
+            if((x1 < lower(1) .or. (x2 < lower(1)) .or. (x1 > upper(1)) .or. (x2 > upper(1)) ) then
+                write(*,*) 'point outside domain was fed into get_dely'
```
matt-frey, line 45: get_dely --> get_delx

src/3d/parcels/parcel_merge.f90
```diff
@@ -160,9 +160,11 @@ subroutine do_group_merge(parcels, isma, iclo, n_merge, Bm, vm)
                 ! temporary scalar containing 1 / vm(m)
                 vmerge = one / vm(m)
 
-                ! x and y centre of merged parcel, modulo periodicity
+                ! x and y centre of merged parcel, modulo periodicity, and apply parcel bc
                 xm(m) = get_delx(x0(m), - vmerge * xm(m))
+                xm(m) = xm(m) - extent(1) * dble(int((xm(m) - center(1)) * hli(1)))
```
matt-frey, line 165: Should we change the code to be able to call the function https://github.com/matt-frey/epic/blob/main/src/parcels/parcel_bc.f90#L17? Or we add that to the parcel_bc.f90 file.

matt-frey (commented):
@sjboeing I just had a brief look.

src/3d/parcels/parcel_merge.f90
matt-frey: @sjboeing  Shouldn't the BC be applied before line 164?

matt-frey: I thought, we either combine `xm` and `ym` to `posm` or add routines to parcel_bc.f90 that can handle single coordinates.

matt-frey: My proposed change is in https://github.com/matt-frey/epic/pull/293.

PR #302, opened 2022-01-21
Change order of multi-dimensional arrays

matt-frey:


matt-frey:
@sjboeing Although I get zero error, I still would like you to have a brief look. Thanks!

sjboeing (approved):
This looks good to me (just went over it quickly, but you already showed results agree).

PR #304, opened 2022-01-22
Remove merger dump

matt-frey:


matt-frey:
@sjboeing Here I remove the merger dump functionality. We only used it during development.

sjboeing (approved):
Looks good to me.

PR #305, opened 2022-01-22
Remove merger dump develop

matt-frey:
This PR applies the changes of #304 to the develop branch.


matt-frey:
@sjboeing This is the same PR like #304 but for the develop branch.

PR #316, opened 2022-02-03
Change setup of parcel correction tests

matt-frey:


unit-tests/2d/test_parcel_correction.f90
```diff
@@ -77,14 +60,50 @@ program test_parcel_correction
...
+
+        tmp = m * val + q
+        parcels%position(1, n) = parcels%position(1, n) + tmp
+
+        call random_number(val)
+        tmp = m * val + q
+        parcels%position(2, n) = parcels%position(2, n) + tmp
+
+        call apply_periodic_bc(parcels%position(:, n))
+        call apply_reflective_bc(parcels%position(:, n), parcels%B(:, n))
```
sjboeing, line 87: Note that this initialisation (with reflective BC) may mean that parcels are less likely to be near a boundary as compared to inside the interior. Not sure if this is the case, would need to give it some thought, or did you think about this already?

sjboeing: It is not a very important issue in any case.

matt-frey: I try to avoid parcel centres being outside the box.

matt-frey: Parcels can still be near the boundary, but they should not be outside. Parcels are only "mirrored" when they are outside the specified domain. If they are just a little bit outside they will end up being just a little bit inside, so close to the boundary.

sjboeing: Ah yes, can see this works out when n_per_dim=2. Happy for you to merge.

sjboeing (approved):

PR #327, opened 2022-02-23
Use parcel vorticity to compute the vorticity tendency

### LINKED ISSUE #245: Revise inversion
As mentioned in #241 there may be some for EPIC unnecessary things happening in the inversion module (see `inversion.f90`). Comments of @daviddritschel:

* Do we always filter after calling lapinv1? If so, could we incorporate the filter into that subroutine and perhaps save an extra loop?$
* Again I wonder if we can incorporate this into diffx ... above, but we do use unfiltered velocity below for advection. Something to come back to: do we really need to do this? In MPIC, we did it because we felt there was not enough small-scale mixing. It might be different in EPIC.
* Should we make as a 4D array like vortg? It would take some careful re-coding to use indices 1,2,3 for the x,y,z components of everything, but it might make the code more readable. But I see that we use a lot of temporary arrays repeatedly, and this may be more memory efficient

matt-frey:
This PR
* removes the solenoidal correction in `vor2vel`
* uses the parcel vorticity to calculate the vorticity tendency
* removes filtering

Closes #245.


src/3d/parcels/parcel_interpl.f90
```diff
@@ -337,8 +343,47 @@ subroutine grid2par(vel, vor, vgrad, add)
...
+                vortend(2, n) =  vortend(2, n)                                   &
+                              +  parcels%vorticity(1, n)           * dvdx        & ! \omegax * dv/dx
+                              + (parcels%vorticity(2, n) + ft_cor) * vgrad(3, n) & ! \omegay * dv/dy
+                              + (parcels%vorticity(3, n) + f_cor)  * dvdz          ! \omegaz * dv/dz
+
+                ! add strain part of vorticity tendency to z-vorticity
+                vortend(3, n) =  vortend(3, n)                                   &
+                              +  parcels%vorticity(1, n)           * vgrad(4, n) & ! \omegax * dw/dx
+                              + (parcels%vorticity(2, n) + ft_cor) * vgrad(5, n) & ! \omegay * dw/dy
+                              - (parcels%vorticity(3, n) + f_cor)  * dwdz          ! \omegaz * dw/dz
```
sjboeing, line 385: I am not sure about the minus sign here. Should it be in the calculation of dwdz?

src/3d/parcels/parcel_interpl.f90
```diff
@@ -337,8 +343,47 @@ subroutine grid2par(vel, vor, vgrad, add)
...
-                    vor(:, n) = vor(:, n) + weights(l) * vtend(ks(l), js(l), is(l), :)
+                    ! du/dz = \omegay + dw/dx
+                    dudz = dudz &
+                         + weights(l) * (vortg(ks(l), js(l), is(l), 2) + velgradg(ks(l), js(l), is(l), 4))
+
+                    ! dv/dz = dw/dy - \omegax
+                    dvdz = dvdz &
+                         + weights(l) * (velgradg(ks(l), js(l), is(l), 5) - vortg(ks(l), js(l), is(l), 1))
+
+                    ! dw/dz = du/dx + dv/dy
```
sjboeing, line 354: See comment below on minus sign.

src/3d/parcels/parcel_interpl.f90
sjboeing: (we may have the right result, but not the right notation)

sjboeing (commented):
Looks good Matthias. I have just put 2 comments in.

src/3d/parcels/parcel_interpl.f90
matt-frey: Yes, the actual notation is in fields.f90, https://github.com/matt-frey/epic/blob/main/src/3d/fields/fields.f90#L31. I will change it here.

src/3d/parcels/parcel_interpl.f90
matt-frey: I fixed the notation.

src/3d/inversion/inversion.f90
```diff
@@ -65,49 +63,13 @@ subroutine vor2vel(vortg,  velog,  velgradg)
             !boundaries (store solution lambda in fs):
             call lapinv1(fs)
```
daviddritschel, line 64: We don't need this anymore since we don't use lambda; note fs is not used after line 66 64 except to store another value

src/3d/inversion/inversion.f90
```diff
@@ -65,49 +63,13 @@ subroutine vor2vel(vortg,  velog,  velgradg)
...
-            !$omp parallel shared(ds, es, fs, as, bs, cs, filt, nz) private(iz) default(none)
-            !$omp do
-            do iz = 0, nz
-                ds(iz, :, :) = filt * as(iz, :, :)
-                es(iz, :, :) = filt * bs(iz, :, :)
-                fs(iz, :, :) = filt * cs(iz, :, :)
-            enddo
-            !$omp end do
-            !$omp end parallel
+            !Compute vorticity in physical space:
```
daviddritschel, line 69: Since we don't correct vortg here, we don't need to recalculate the vorticity in physical space - it is still just vortg.  There are major simplifications to be made still in this code

src/3d/inversion/inversion.f90
```diff
@@ -65,49 +63,13 @@ subroutine vor2vel(vortg,  velog,  velgradg)
...
+            ds = as
+            es = bs
+            fs = cs
 
             !Save boundary values of x and y vorticity for z derivatives of A & B:
```
daviddritschel, line 74: We do need to do this in order to get the velocity field from inverting the spectral vorticity

src/3d/inversion/inversion.f90
daviddritschel: We don't need lines 134-137 below as a result

daviddritschel (commented):
There are major simplifications still to be made - won't change the results I think

src/3d/inversion/inversion.f90
matt-frey: True. I missed the line `fs = cs`. I just saw that `fs` was used, but missed that it is overwritten.

src/3d/inversion/inversion.f90
matt-frey: `vortg` is `inout`, we need to copy in order to avoid inverse FFT.

matt-frey: Not sure what you mean with lines 34-137, we need to calculate the velocity in spectral space since we use in the velocity gradient calculation called within vor2vel.

src/3d/inversion/inversion.f90
matt-frey: I removed these assignments.

src/3d/inversion/inversion.f90
```diff
@@ -34,43 +35,18 @@ subroutine vor2vel(vortg,  velog,  velgradg)
 
             call start_timer(vor2vel_timer)
 
+            ! Copy vorticity vector field
+            vortg_copy = vortg
```
daviddritschel, line 39: You don't need a copy since the fftxyp2s and fftxys2p do not modify the input array.

src/3d/inversion/inversion.f90
```diff
@@ -34,43 +35,18 @@ subroutine vor2vel(vortg,  velog,  velgradg)
...
-            !$omp end workshare
-            !$omp end parallel
-
-            !Remove horizontally-averaged part (plays no role):
-            fs(:, 0, 0) = zero
-
-            !Invert Lap(lambda) = div(vortg) assuming dlambda/dz = 0 at the
-            !boundaries (store solution lambda in fs):
-            call lapinv1(fs)
+            call fftxyp2s(vortg_copy(0:nz, :, :, 1), as)
```
daviddritschel, line 43: Just use vortg here

daviddritschel: Also below, can you just send in all components of svelog into the FFT routine, or is that routine expecting a 3D array of a certain structure?  I think the latter is true.

daviddritschel (commented):
A simplification I think - you don't need to copy vortg.

src/3d/inversion/inversion.f90
matt-frey: As far as I see `fftxyp2s` do change the physical space input.

src/3d/inversion/inversion.f90
matt-frey: Yes, it expects a 3D array.

src/3d/inversion/inversion.f90
matt-frey: I dug deeper and found out that some FFT routines take arguments as `inout` although it is just `in`. I will change this, so I guess, we really do not need to copy!!

matt-frey: We need to change the signature of `forfft(m, n, x, trig, factors)` to avoid the copy of vortg. The input array `x` is modified.

matt-frey: Instead of having an additional field, we can use `velog`; hence, we do not need additional space.

src/3d/inversion/inversion.f90
```diff
@@ -26,8 +26,7 @@ subroutine vor2vel(vortg,  velog,  velgradg)
             double precision                :: ds(0:nz, 0:nx-1, 0:ny-1) &
                                              , es(0:nz, 0:nx-1, 0:ny-1) &
                                              , fs(0:nz, 0:nx-1, 0:ny-1)
```
daviddritschel, line 28: We no longer need fs I think

src/3d/inversion/inversion_utils.f90
```diff
@@ -141,6 +146,9 @@ subroutine init_fft
                 enddo
             enddo
 
+            k2l2i(1, 1) = zero ! k = l = 0
+            k2l2i(2:, 2:) = one / ksq(2:, 2:)
```
daviddritschel, line 150: Does this define the 2,1 and the 1,2 elements?

daviddritschel (commented):
Couple of issues to look at

src/3d/inversion/inversion_utils.f90
matt-frey: `2:` means accessing indices 2, 3, ...

matt-frey: The array starts at 1 not at zero.

src/3d/inversion/inversion.f90
matt-frey: Indeed. I missed that. Thanks.

matt-frey: I removed.

src/3d/inversion/inversion_utils.f90
matt-frey: Oh, yes!!

matt-frey: I fill in a loop. Is probably safest.

matt-frey: Hmm, we also need to change the order of dimensions for k2l2i.

matt-frey: Ah, no. The order is fine.

matt-frey: Fixed.

daviddritschel (approved):

matt-frey:
Thanks for approving. I pull this into main. I created a separate branch for the vor2vel version with filtering and solenoidal correction.

PR #329, opened 2022-03-09
Add python EPIC config and write netCDF header

matt-frey:


matt-frey:
The only thing I dislike at the moment is the hard-coding of CF-1.8 at 2 locations.

PR #332, opened 2022-03-10
Fix reading netcdf datasets

matt-frey:


sjboeing (commented):
Thanks, Matthias, this looks good to me!

PR #334, opened 2022-03-11
Dynamic maximum number of parcels

matt-frey:
We currently set the maximum number of parcels as a constant `max_num_parcels`. This causes issues when increasing the grid resolution. We should make this a runtime dependent value. I suggest to use `max_num_parcels = nx * ny * nz * n_per_cell * size_factor` (and `nx * nz * n_per_cell * size_factor` in 2D) where `size_factor` is a user-defined input parameter.

### DIFF
--- a/src/2d/utils/parameters.f90
+++ b/src/2d/utils/parameters.f90
@@ -57,6 +57,9 @@ module parameters
     ! maximum volume
     double precision :: vmax
 
+    ! maximum number of allowed parcels
+    integer :: max_num_parcels
+
     contains
 
     ! Update all parameters according to the
@@ -97,6 +100,8 @@ module parameters
 
         vmin = vcell / parcel%min_vratio
         vmax = vcell / parcel%max_vratio
+
+        max_num_parcels = nx * nz * parcel%n_per_cell * parcel%size_factor
 
     end subroutine update_parameters
 end module parameters
--- a/src/2d/utils/parameters.f90
+++ b/src/2d/utils/parameters.f90
@@ -101,7 +101,7 @@ module parameters
         vmin = vcell / parcel%min_vratio
         vmax = vcell / parcel%max_vratio
 
-        max_num_parcels = nx * nz * parcel%n_per_cell * parcel%size_factor
+        max_num_parcels = int(nx * nz * parcel%min_vratio * parcel%size_factor)
 
     end subroutine update_parameters
 end module parameters
--- a/src/3d/parcels/parcel_container.f90
+++ b/src/3d/parcels/parcel_container.f90
@@ -5,6 +5,7 @@
 module parcel_container
     use options, only : verbose
     use parameters, only : extent, hli, center, lower, upper
+    use parcel_ellipsoid, only : parcel_ellispoid_allocate, parcel_ellispoid_deallocate
     implicit none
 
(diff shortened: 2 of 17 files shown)


matt-frey:
@sjboeing I would be happy you could have a look at this PR. I checked if I can run EPIC and restart in 2D and 3D as well.

src/3d/parcels/parcel_container.f90
```diff
@@ -5,6 +5,7 @@
 module parcel_container
     use options, only : verbose
     use parameters, only : extent, hli, center, lower, upper
+    use parcel_ellipsoid, only : parcel_ellispoid_allocate, parcel_ellispoid_deallocate
```
sjboeing, line 8: Note the spelling: ellipsoid (throughout).

src/2d/utils/parameters.f90
```diff
@@ -98,5 +101,7 @@ subroutine update_parameters
         vmin = vcell / parcel%min_vratio
         vmax = vcell / parcel%max_vratio
 
+        max_num_parcels = nx * nz * parcel%n_per_cell * parcel%size_factor
```
sjboeing, line 104: Maybe it is better to relate this to the maximum number of parcels per grid box/minimum parcel size instead (in that case, a factor 1 would be completely safe, but in practice we may get away with something like 0.5).

sjboeing:
This seems like a sensible approach (maybe we could check with Evegenij). I left a few comments. Did you check restarts from both parcels and grid?

src/3d/parcels/parcel_container.f90
matt-frey: Thanks, I didn't see this typo.

matt-frey:
Yes, I checked restarts from parcels and grids.

src/2d/utils/parameters.f90
matt-frey: Do you mean something like `max_num_parcels = nx * nz * vcell / vmin`?

sjboeing: Yes, but still with a size factor, I think, because typically we can get away with fewer parcels if needed.

matt-frey: Ok, but since you suggested 0.5 above, I will change the `size_factor` to a float. When I added this parameter I also thought that a float would be a reasonable choice.

sjboeing: Yes, that seems right. Maybe we won't always manage with 0.5, 1 should be relatively safe. The TG simulations will give you an indication for a reasonable value. The eventual max_num_parcels would be an int.

matt-frey: I put 1 as default for the moment. I only changed the type of `size_factor`, `max_num_parcels` is still an int.

matt-frey: See my latest commit.

matt-frey: The explicit type-cast around the expression is in principle not necessary.

matt-frey:
@sjboeing Can I pull this change into main?

sjboeing:
@matt-frey: yes, happy for you to merge this!

matt-frey:
Great! Thanks!

PR #342, opened 2022-03-23
Fix calculation of symmetrized strain matrix

matt-frey:


src/3d/stepper/rk4_utils.f90
```diff
@@ -115,12 +115,23 @@ function get_time_step(t) result(dt)
...
                         !    dw/dz = - (du/dx + dv/dy)
-                        strain(1, 1) = velgradg(iz, iy, ix, 1)                              ! du/dx
-                        strain(1, 2) = two * velgradg(iz, iy, ix, 2) + vortg(iz, iy, ix, 3) ! du/dy + dv/dx
-                        strain(1, 3) = two * velgradg(iz, iy, ix, 4) + vortg(iz, iy, ix, 2) ! du/dz + dw/dx
-                        strain(2, 2) = velgradg(iz, iy, ix, 3)                              ! dv/dy
-                        strain(2, 3) = two * velgradg(iz, iy, ix, 5) - vortg(iz, iy, ix, 1) ! dv/dz + dw/dy
-                        strain(3, 3) = -(velgradg(iz, iy, ix, 1) + velgradg(iz, iy, ix, 3)) ! dw/dz
+                        !
+                        !                         /  2 * u_x  u_y + v_x u_z + w_x\
+                        ! 1/2 * (S + S^T) = 1/2 = |u_y + v_x   2 * v_y  v_z + w_y|
```
daviddritschel, line 120: 1/2 = |   should be.  1/2 * |   here I think

daviddritschel (commented):
Just a text comment

src/3d/stepper/rk4_utils.f90
matt-frey: Indeed.

PR #344, opened 2022-03-24
Change to criterion used in paper

matt-frey:
In the current version of the 3D paper we write the Jacobi convergence criterion is when the absolute sum of the off-diagonals `sm` is less than `atol`. We need to make the exit statement of the `while` loop to be `sm >= atol` to be consistent.


sjboeing (approved):
Happy for you to change this. It should only change the result in the very rare case that sm==atol, as I understand.

matt-frey:
Yes.

PR #351, opened 2022-03-30
Add with correct sign

matt-frey:
We need to add epsilon with the correct sign, we can get division by zero otherwise.

### DIFF
--- a/src/utils/jacobi.f90
+++ b/src/utils/jacobi.f90
@@ -26,12 +26,12 @@ module jacobi
             g = hundred * dabs(aij)
             h = D(j) - D(i)
             if (dabs(h) + g == dabs(h)) then
-                ! sign(A, B) returns B with the sign of A
+                ! sign(A, B) returns A with the sign of B
                 ! We need to add epsilon with the proper sign
                 ! since h might be -epsilon causing a division by zero.
                 t = aij / (h + sign(epsilon(h), h))
             else
-                ! sign(A, B) returns B with the sign of A
+                ! sign(A, B) returns A with the sign of B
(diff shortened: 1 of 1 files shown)


src/utils/jacobi.f90
```diff
@@ -26,9 +26,15 @@ pure subroutine givens(A, D, i, j, c, s, t, tau)
...
+                ! We need to add epsilon with the proper sign
+                ! since h might be -epsilon causing a division by zero.
+                t = aij / (h + sign(epsilon(h), h))
             else
-                theta = f12 * h / (aij + epsilon(aij))
+                ! sign(A, B) returns A with the sign of B
+                ! We need to add epsilon with the proper sign
+                ! since aij might be -epsilon causing a division by zero.
+                theta = f12 * h / (aij + sign(epsilon(aij), aij))
                 t = one / (dabs(theta) + dsqrt(one + theta ** 2))
```
daviddritschel, line 38: Could we compute t from

t = aij / (f12 * h + dsqrt(aij**2 + (f12*h)**2)

to avoid having to add this epsilon?  Or is this less efficient?

daviddritschel (commented):
Just a comment about a calculation ; I guess my suggestion fails if BOTH aij and h are zero

src/utils/jacobi.f90
matt-frey: theta is used later as well, that is why it is split up.

daviddritschel: OK, I am happy so long as it is sufficiently efficient (a mouthful of words!).  Thanks for spotting these corrections.

PR #360, opened 2022-04-13
Use relative error

matt-frey:


sjboeing (approved):
I think this makes sense, although we now have a division rather than subtraction, which in some way may be less precise.

PR #361, opened 2022-04-14
Field decomposition

matt-frey:


matt-frey:
Hi @jevbelikov and @sjboeing, I think this PR is ready for review. We might still change a few things later, but I think it is a first step into the right direction.

src/3d/fields/field_layout.f90
```diff
@@ -0,0 +1,133 @@
...
+        integer :: xlo, xhi
+        integer :: ylo, yhi
+        integer :: corners(4)
+    end type neighbour_type
+
+    type(box_type)       :: box
+    type(neighbour_type) :: neighbour
+
+    ! we always have 1 internal halo layer
+    integer :: n_halo = 1
```
sjboeing, line 21: I think the different meanings of nh and n_halo are a bit confusing.

src/3d/fields/fields.f90
```diff
@@ -48,31 +49,38 @@ module fields
...
+            call field_layout_init(nx, ny, nz, nh)
+
+            lo = box%hlo(3)
+            hi = box%hhi(3)
+
+            allocate(velog(lo(3):hi(3), lo(2):hi(2), lo(1):hi(1), 3))
+            allocate(velgradg(lo(3):hi(3), lo(2):hi(2), lo(1):hi(1), 5))
 
-            allocate(volg(-1:nz+1, 0:ny-1, 0:nx-1))
+            allocate(volg(lo(3):hi(3), lo(2):hi(2), lo(1):hi(1)))
```
sjboeing, line 66: Do these fields not include the halo?

unit-tests/mpi/test_mpi_layout.f90
```diff
@@ -0,0 +1,51 @@
...
+!
+!               This unit test checks the MPI init and finalize.
+! =============================================================================
+program test_mpi_layout
+    use unit_test
+    use mpi_communicator
+    use field_layout
+    implicit none
+
+    integer, parameter            :: nx = 32, ny = 32, nz = 32
```
sjboeing, line 12: Should these have default values?

src/3d/fields/fields.f90
sjboeing: Ah, I see lo is box%hlo. I think it is tricky to use the same name that is used for the internal boundaries elsewhere.

matt-frey: They should. Yeah, I should maybe use another name.

src/3d/fields/field_layout.f90
matt-frey: I removed the variable since it is not used at the moment.

PR #362, opened 2022-04-15
Parcel exchange

matt-frey:
This PR is a first version for the parcel halo swap. It still needs some changes in the MPI communication though.


matt-frey:
@sjboeing @jevbelikov A first version of the parcel halo swap is implemented. We must improve the MPI communication in a next step, but I guess this can be a starting point.

PR #365, opened 2022-04-18
Read and write parcels in parallel

matt-frey:
We currently only support reading parcels if the start index dataset is part of the provided NetCDF parcel file. See also #371.

### DIFF
--- a/src/3d/parcels/parcel_netcdf.f90
+++ b/src/3d/parcels/parcel_netcdf.f90
@@ -320,7 +320,7 @@ module parcel_netcdf
             remaining = n_total_parcels - n_parcels * mpi_size
             start_index = n_parcels * mpi_rank
 
-            if (mpi_rank < remaining)
+            if (mpi_rank < remaining) then
                 n_parcels = n_parcels + 1
             endif
 
--- a/src/3d/parcels/parcel_netcdf.f90
+++ b/src/3d/parcels/parcel_netcdf.f90
@@ -324,6 +324,8 @@ module parcel_netcdf
                 n_parcels = n_parcels + 1
             endif
 
+            ! all MPI ranks < remaining get an additional parcel;
+            ! hence, we must shift the start indices accordingly
             start_index = start_index + min(remaining, mpi_rank)
 
             if (n_parcels > max_num_parcels) then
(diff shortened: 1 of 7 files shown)


src/3d/parcels/parcel_netcdf.f90
```diff
@@ -255,9 +260,19 @@ subroutine write_netcdf_parcels(t)
             ! write time
             call write_netcdf_scalar(ncid, t_axis_id, t, 1)
 
-            ! time step to write [step(2) is the time]
-            cnt   = (/ n_parcels, 1 /)
-            start = (/ 1,         1 /)
+            ! after this operation all MPI ranks now their starting index
```
sjboeing, line 263: now -> know (I regularly do this myself)

sjboeing: There are some advantages to putting the parcels in separate files. In fact, the subgrid visualisation algorithm currently exploits that parcels are in separate files, as it processes the subdomains using shared-memory parallelism: this has the advantage that only a small part of the full grid is used at once). Alternatively, each parcel could write its processor id for processing.

sjboeing: I was also wondering what works better for restarts (if it is on the same decomposition, separate files could be easier). Maybe Evgenij would know, especially for large core counts.

sjboeing (commented):
Looks good, Matthias! Would be good to ask Evgenij what he thinks is the best solution (a single file or multiple files). For PMPIC, we have used one file per processor, which can have advantages for post-processing.

src/3d/parcels/parcel_netcdf.f90
matt-frey: Writing many "small" files is usually not advantageous. With high core counts you may also overload the file system. Didn't you experience something like this for a huge PMPIC run?

matt-frey: Post-processing can happen in parallel as well although it is a single file.

sjboeing: The post-processing exploits the fact that each file only contains a small subdomain.

sjboeing: Each thread has a copy of the grid, but because each file contains only part of the full grid, these copies only contain part of the grid. There is a single copy of the full grid.

sjboeing: But yes, we eventually ran into an overloaded file system at a 1024**3 grid, so that may be better with a single file. Maybe writing a "grid id" as one of the parcel properties will mean we can have the best of both.

PR #377, opened 2022-05-02
Parcel rejection method

### LINKED ISSUE #371: Fix reading parcels if the start index dataset is not provided
We currently only support reading parcels with MPI if the start index dataset is provided. In case this is missing there are two ways of approaching this:
* read by rejection (i.e. read only parcels that are inside the sub-domain owned by the MPI rank)
* read a subset and then redistribute parcels according to the sub-domains

matt-frey:
Closes #371.

### DIFF
--- a/src/3d/fields/fields.f90
+++ b/src/3d/fields/fields.f90
@@ -115,6 +115,19 @@ module fields
             k = floor((pos(3) - lower(3)) * dxi(3))
         end subroutine get_index
 
+        pure function is_contained(pos) result(l_contained)
+            double precision, intent(in) :: pos(3)
+            integer                      :: i, j, k
+            logical                      :: l_contained
+
+            call get_index(pos, i, j, k)
+
+            l_contained = ((i > box%hlo(1))   .and. &
+                           (i < box%hhi(1)-1) .and. &
+                           (j > box%hlo(2))   .and. &
+                           (j < box%hhi(2)-1))
+        end function
+
 
         ! Do periodic shift of the index
         ! @param[inout] ii zonal grid point indices
--- a/src/3d/fields/fields.f90
+++ b/src/3d/fields/fields.f90
@@ -122,10 +122,10 @@ module fields
 
             call get_index(pos, i, j, k)
 
-            l_contained = ((i > box%hlo(1))   .and. &
-                           (i < box%hhi(1)-1) .and. &
-                           (j > box%hlo(2))   .and. &
-                           (j < box%hhi(2)-1))
+            l_contained = ((i >= box%lo(1))  .and. &
+                           (i <= box%hi(1))  .and. &
+                           (j >= box%lo(2))  .and. &
+                           (j <= box%hi(2)))
         end function
 
 
--- a/src/3d/parcels/parcel_netcdf.f90
+++ b/src/3d/parcels/parcel_netcdf.f90
@@ -6,10 +6,7 @@ module parcel_netcdf
     use parcel_container, only : parcels            &
                                , n_parcels          &
                                , n_total_parcels    &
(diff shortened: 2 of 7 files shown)


src/3d/fields/fields.f90
```diff
@@ -115,6 +115,19 @@ pure subroutine get_index(pos, i, j, k)
...
         end subroutine get_index
 
+        pure function is_contained(pos) result(l_contained)
+            double precision, intent(in) :: pos(3)
+            integer                      :: i, j, k
+            logical                      :: l_contained
+
+            call get_index(pos, i, j, k)
+
+            l_contained = ((i > box%hlo(1))   .and. &
```
sjboeing, line 125: This is assuming halos of size 1, right.

src/3d/parcels/parcel_netcdf.f90
```diff
@@ -355,25 +365,18 @@ subroutine read_netcdf_parcels(fname)
...
-                endif
-
-                ! all MPI ranks < remaining get an additional parcel;
-                ! hence, we must shift the start indices accordingly
-                start_index = start_index + min(remaining, mpi_rank)
-
-                ! FIXME
-                call mpi_exit_on_error("This is not yet supported! See issue #371")
-
+                call mpi_print("WARNING: The start index is not provided. All MPI ranks read all parcels!")
```
sjboeing, line 368: It could help to read variable by variable in such cases, or not read everything into memory directly in a different way (read in chunks?). could be left for future work.

sjboeing (commented):
Thanks Matthias, I was wondering about reading in chunk-by-chunk in such cases. Can be left for later.

src/3d/fields/fields.f90
matt-frey: Yes, but we will keep it fix, I thought.

matt-frey:
Yes, I haven't thought about this actually.

matt-frey:
@sjboeing I changed the code such that we read chunk by chunk.

src/3d/parcels/parcel_netcdf.f90
matt-frey: Done.

src/3d/fields/fields.f90
sjboeing: I think so, if we link to hi and lo here we are more flexible when we change it, but I don't think we will.

matt-frey: Yes, the current version restricts to a single halo in the lower boundary. Agreed, I can change to make it more general in order to prevent future errors.

src/3d/parcels/parcel_netcdf.f90
```diff
@@ -354,35 +360,82 @@ subroutine read_netcdf_parcels(fname)
...
-            ! since the starting index in Fortran is 1 and not 0.
-            start = (/ 1 + start_index, 1 /)
-            cnt   = (/ n_parcels,       1 /)
+            call stop_timer(parcel_io_timer)
+
+        end subroutine read_netcdf_parcels
+
+
+        ! This subroutine assumes the NetCDF file to be open.
+        subroutine read_chunk(first, last)
```
matt-frey, line 431: The reading in case of chunks is wrong, the range must be changed in the parcel container.

src/3d/fields/fields.f90
matt-frey: Done. 40a20db2a55714f929b8874f0642c7cb216b82a4

src/3d/parcels/parcel_netcdf.f90
matt-frey: It works now.

PR #380, opened 2022-05-04
Add split merge unit test

matt-frey:


sjboeing (approved):
Thanks Matthias, this looks sensible. Good to try different angles.

PR #403, opened 2022-09-13
Apply nearest fix in 2d

matt-frey:
This does the same fix in 2D as in 3D. However, we did not experience any issues in 2D.


sjboeing (approved):
I support bringing these changes in. Looks good to me!

PR #406, opened 2022-09-21
Generalise NetCDF dimension and axis names

matt-frey:
This PR sightly generalises the NetCDF interface as the shallow water code is fully periodic and has coordinate names 'x' and 'y' for horizontal and vertical rather than 'x' and 'z' as in EPIC2D.


matt-frey:
@sjboeing After our discussion, I improved the interface, that is, I added two subroutines to set the dimension and axis names.

sjboeing (approved):
Apologies I did not get back to this before, looks good!

PR #407, opened 2022-09-27
Monitor number of splits and merges

matt-frey:


sjboeing (commented):
@matt-frey: thanks for your work on this. Maybe it is quickest to go over this after/in the meeting today to make sure I understand. Just checking: the number of time steps between diagnostic writes is available already?

matt-frey:
@sjboeing I simplified the interface. We now have a single variable each.

sjboeing (commented):
Hi Matthias, thanks for writing the diagnostics. I was wondering how the splits are now currently added between diagnostic writes, or is it now just the last time step? Sorry, may have taken a too quick look at it!

sjboeing:
What I thought was to have a variable which is only modified in the splitting/merging routine that keeps track of the number in the individual time step, and another one (the sum) 
which is incremented in these routines and reset after the diagnostic writes.

matt-frey:
Okay, but if we do not dump the individual time steps, I do not see a purpose of the additional variables.

matt-frey:
If you are okay, we may dump the individual number of splits and writes in debug mode (and or verbose) to an ASCII file.

sjboeing:
Ah, maybe best to have a quick call so I understand correctly what is happening!

sjboeing (approved):
Ah, I see now, sorry, got confused! Happy for you to merge this.

PR #408, opened 2022-09-27
Zeta zero on boundary

matt-frey:


src/3d/utils/parameters.f90
```diff
@@ -114,4 +127,40 @@ subroutine update_parameters
...
+    subroutine set_zeta_boundary_flag(zeta, thres)
+        double precision, intent(in) :: zeta(-1:nz+1, 0:ny-1, 0:nx-1)
+        double precision, intent(in) :: thres
+        double precision :: rms_bndry(2), rms_interior
+
+        rms_interior = dsqrt(sum(zeta(1:nz-1, :, :) ** 2) * nhcelli / (nz-1))
+        rms_bndry(1) = dsqrt(sum(zeta(0,      :, :) ** 2) * nhcelli)
+        rms_bndry(2) = dsqrt(sum(zeta(nz,     :, :) ** 2) * nhcelli)
+
+        l_bndry_zeta_zero(:) = (rms_bndry(:) < thres * (epsilon(rms_interior) + rms_interior))
```
sjboeing, line 141: Is "thres*rms_interior+epsilon(rms_interior)" better?

matt-frey: Yes, we can also use this criterion. Maybe better as with the other eq. you need a large thres value.

src/3d/utils/parameters.f90
```diff
@@ -114,4 +130,81 @@ subroutine update_parameters
...
+
+
+    subroutine write_zeta_boundary_flag(ncid)
+        integer, intent(in)     :: ncid
+        integer                 :: grp_ncid
+        character(*), parameter :: name = 'parameters'
+
+        ncerr = nf90_def_grp(ncid, name, grp_ncid)
+
+        call check_netcdf_error("Faild to create NetCDF group '" // name // "'.")
```
sjboeing, line 186: Failed

matt-frey: Thanks ... I make many English typos.

matt-frey:
I thought we justified our approach. @daviddritschel What do you think?

sjboeing:
(if it needs changing, this could still be done in a later PR in any case)

matt-frey:
I prefer having a final good solution now. We can add an option if necessary that forces zero on the boundary.

PR #410, opened 2022-11-07
Fix compilation with config.f90

matt-frey:


sjboeing (approved):
Thanks Matthias: this compiles for me. Happy to merge this.

PR #425, opened 2022-11-21
Use central differencing in solenoidal correction and tendency calculation

matt-frey:
This PR
- use FD for performing vertical derivatives
- fixes the subroutine diffz
- adds more unit tests for diffz and central_diffz

### DIFF
--- a/src/3d/inversion/inversion.f90
+++ b/src/3d/inversion/inversion.f90
@@ -51,8 +51,7 @@ module inversion_mod
             ds = as - bs                     ! ds = D
             cs = svor(:, :, :, I_Z)
             !$omp end parallel workshare
-!             call diffz(cs, es)               ! es = E
-call field_combine_semi_spectral(cs)
+            call field_combine_semi_spectral(cs)
             call central_diffz(cs, es)                     ! es = E
             call field_decompose_semi_spectral(es)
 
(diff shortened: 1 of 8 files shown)


src/3d/inversion/inversion.f90
```diff
@@ -342,23 +344,18 @@ subroutine divergence(f, div)
             ! calculate df1/dx
             call fftxyp2s(f(0:nz, :, :, I_X), fs)
             call diffx(fs, ds)
-            call fftxys2p(ds, div)
+            call fftxys2p(ds, f(0:nz, :, :, I_X))
```
daviddritschel, line 347: Is it ok that we overwrite f here; is it not needed in the calling routine?

src/3d/inversion/inversion.f90
```diff
@@ -342,23 +344,18 @@ subroutine divergence(f, div)
...
 
             ! calculate df2/dy
             call fftxyp2s(f(0:nz, :, :, I_Y), fs)
             call diffy(fs, ds)
             call fftxys2p(ds, f(0:nz, :, :, I_Y))
```
daviddritschel, line 352: Same here

daviddritschel (commented):
Just a couple of minor questions - can commit if they are not a concern

src/3d/inversion/inversion.f90
matt-frey: It is fine. The call to `fftxyp2s` which convertes f to semi-spectral spce (stored in `fs`) also overwrites the array `f`.

matt-frey: The "divergence" routine also has a comment saying "Note: f is overwritten".

src/3d/inversion/inversion.f90
matt-frey: See my other reply.

PR #430, opened 2022-12-02
Check if file exists

### LINKED ISSUE #429: Issue error if input file not present.
Just noticed that when there is no input file, it says:
 ****************************
  Factorisation not possible.
  Only factors from 2-6 allowed.
  STOPPING...
 ****************************
Possibly we can do an additional check on the file existing, and then spit our an error.

matt-frey:
Closes #429.


sjboeing (approved):
Thanks, Matthias, this fixes the issue.

PR #449, opened 2023-04-06
Remove reference potential energy calculation for 3D

### LINKED ISSUE #370: Remove the calculation of the reference potential energy in 3D
In order to calculate the reference potential energy (`peref`) the buoyancy must be ordered globally. This is not easily done with MPI. The parcel diagnostics in #368 does not take this into account at the moment and hence, the (reference) potential energy is wrong. This issue must also address the missing potential energy check in the unit test `test_mpi_parcel_diagnostics`.

matt-frey:
We remove the peref calculation for 3D as the result is too inaccurate and the calculation involves a global sort across all MPI ranks which we do not want to do.

Closes #370.

### DIFF
--- a/src/3d/parcels/parcel_diagnostics.f90
+++ b/src/3d/parcels/parcel_diagnostics.f90
@@ -3,14 +3,13 @@
 ! =============================================================================
 module parcel_diagnostics
     use constants, only : zero, one, f12, thousand
-    use merge_sort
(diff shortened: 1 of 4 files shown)


sjboeing (approved):
Looks good to me, Matthias! Thank you for doing this.

matt-frey:
@sjboeing Thanks!

PR #451, opened 2023-04-14
Add boundary surface fluxes

matt-frey:
Add boundary surface fluxes for buoyancy and humidity as in MPIC.


src/3d/fields/fields.f90
```diff
@@ -118,6 +119,19 @@ pure subroutine get_index(pos, i, j, k)
             k = floor((pos(I_Z) - lower(I_Z)) * dxi(I_Z))
         end subroutine get_index
 
+        ! Get the lower horizontal (x, y) index of the cell the parcel is in.
+        ! This subroutine does not take x periodicity into account.
+        ! @param[in] pos position of the parcel
+        ! @param[out] i lower, zonal cell index
+        ! @param[out] j lower, vertical cell index
```
sjboeing, line 126: vertical -> meridional?

src/3d/parcels/parcel_interpl.f90
```diff
@@ -473,4 +473,45 @@ pure subroutine get_weights(xyz, i, j, k, ww)
 
         end subroutine get_weights
 
+        !::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+
+        ! Bi-linear interpolation
+        ! @param[in] pos position of the parcel
+        ! @param[out] ii horizontal grid points for interoplation
+        ! @param[out] jj vertical grid points for interpolation
```
sjboeing, line 481: vertical -> meridional?

src/3d/boundary_layer/bndry_fluxes.f90
```diff
@@ -0,0 +1,289 @@
...
+                return
+            endif
+
+            call open_netcdf_file(fname, NF90_NOWRITE, ncid)
+
+            lo = box%lo
+            hi = box%hi
+
+            ! need to add 1 since start must begin with index 1
+            start(1:2) = lo(1:2) + 1
```
sjboeing, line 129: Thanks, I find this a little confusing (would think lo is 1 as well) but must be right.

matt-frey: The indices of our boxes start with 0. According to the [NetCDF documentation](https://docs.unidata.ucar.edu/netcdf-fortran/current/f90-variables.html#f90-reading-data-values-nf90_get_var) the variable `start` must begin with 1, unless I misunderstand.

matt-frey: We must therefore shift each index by 1.

sjboeing: Zero-indexing is sensible for EPIC.

sjboeing (commented):
Looks good to me. The one thing that would/will take me some more time is to carefully check the halo indices, though maybe the unit tests largely take care of this.

matt-frey:
Which halo indices?

matt-frey:
Yes, I copied the 3D ones and removed the vertical direction.

sjboeing (approved):
This looks good to me, Matthias. Let me know if there is anything specific that I should look into. Are there any tests that would be worth putting in with surface fluxes (e.g. conservation)?

matt-frey:
Do you think of unit tests? You might have a look at the time step estimate. I put in a estimate to ensure that initial time steps include the boundary fluxes.

PR #466, opened 2023-05-02
Cleanup and additional debug check in parcel nearest

matt-frey:
This PR
* removes print statements used for debugging
* adds an additional check in debug mode


sjboeing (approved):
Looks good!

PR #469, opened 2023-05-03
Print on rank 0 only and use MPI_Abort instead of stop

matt-frey:
This PR
* replaces `print` statements with `mpi_print` such that only MPI rank 0 writes to the terminal.
* adds a new subroutine `mpi_stop` to exit the program without error.
* replaces occurrences of `MPI_Abort` with `mpi_exit_on_error`.


src/3d/fields/field_netcdf.f90
```diff
@@ -430,8 +431,7 @@ subroutine read_netcdf_fields(fname, step)
             if (st == -1) then
                 st = n_steps
             else if ((st == 0) .or. (st > n_steps)) then
-                print *, "Step number out of bounds."
-                error stop
+                call mpi_stop("Step number in NetCDF field file is out of bounds.")
```
sjboeing, line 434: Here, mpi_stop has a text argument. Is this different from mpi_print+mpi_stop above?

src/3d/stepper/rk4_utils.f90
```diff
@@ -204,7 +205,7 @@ function get_time_step(t) result(dt)
 
             if (dt <= zero) then
                 print "(a10, f0.2, a6)", "Time step ", dt, " <= 0!"
```
sjboeing, line 207: Should this be an mpi_print?

sjboeing (commented):
Thanks Matthias. Happy for you to merge this if you have done a test compilation+run.

src/3d/fields/field_netcdf.f90
matt-frey: I tried to replace all occurrences of `mpi_print` followed by `mpi_stop` with an `mpi_stop` only that takes a text argument.

src/3d/stepper/rk4_utils.f90
matt-frey: That is the issue I explained in Slack. I cannot mimic the print signature at the moment. That is why I have to do it here like this.

matt-frey:
Yes, I also explicitly checked if I can still run the 2D model properly.

PR #470, opened 2023-05-04
Communicate multiple fields in a single call

matt-frey:
We currently call communication routines for each field (and vector field component) individually. This PR changes to single calls where the communication buffers are allocated to fit multiple fields.


src/3d/fields/field_mpi.f90
```diff
@@ -92,11 +246,41 @@ subroutine field_halo_swap(data)
...
 
-        end subroutine field_halo_swap
+            call field_mpi_dealloc
+
+        end subroutine field_halo_swap_scalar
+
+        !::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+
+        subroutine field_halo_swap_vector(data)
+            double precision, intent(inout) :: data(box%hlo(3):, &
```
sjboeing, line 262: No need for end indices?

src/3d/parcels/parcel_correction.f90
```diff
@@ -147,16 +145,11 @@ subroutine apply_laplace(l_reuse)
...
             ! use extrapolation to fill z grid lines outside domain:
-            ! (anti-symmetry in w is equal to extrapolation)
-            ud(-1, :, :)   =  two * ud(0, :, :) - ud(1, :, :)
-            vd(-1, :, :)   =  two * vd(0, :, :) - vd(1, :, :)
-            wd(-1, :, :)   = -wd(1, :, :)
-            ud(nz+1, :, :) =  two * ud(nz, :, :) - ud(nz-1, :, :)
-            vd(nz+1, :, :) =  two * vd(nz, :, :) - vd(nz-1, :, :)
-            wd(nz+1, :, :) = -wd(nz-1, :, :)
+            grad(-1,   :, :, :) =  two * grad(0,  :, :, :) - grad(1,    :, :, :)
+            grad(nz+1, :, :, :) =  two * grad(nz, :, :, :) - grad(nz-1, :, :, :)
```
sjboeing, line 152: This is not the same BC for wd?

sjboeing: Ah, OK, maybe no problem since wd(0, ...)=0.

src/3d/parcels/parcel_init.f90
```diff
@@ -206,4 +204,35 @@ subroutine init_parcels_from_grids
 
         end subroutine init_parcels_from_grids
 
+        !::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+
+        subroutine init_fill_halo
+            integer :: n_fields = 4
```
sjboeing, line 210: Could this be a parameter if coded differently?

sjboeing: Like below

src/3d/parcels/parcel_correction.f90
matt-frey: Yes, since w(0, ...) = w(nz, ...) = 0, anti-symmetry is extrapolation.

src/3d/parcels/parcel_init.f90
matt-frey: Can be a parameter.

src/3d/fields/field_mpi.f90
matt-frey: No, this is not needed.

src/3d/fields/field_mpi.f90
```diff
@@ -5,38 +5,140 @@ module field_mpi
...
+
+        subroutine field_halo_fill_scalar(data, l_alloc)
+            double precision,  intent(inout) :: data(box%hlo(3):box%hhi(3), &
+                                                     box%hlo(2):box%hhi(2), &
+                                                     box%hlo(1):box%hhi(1))
+            logical, optional, intent(in)    :: l_alloc
+            logical                          :: l_buf
+
+            l_buf = .true.
+            if (present(l_alloc)) then
```
sjboeing, line 81: So this is true unless explicitly specified as false. This is OK, though wondering if making it a mandatory flag would be easier.

src/3d/fields/field_mpi.f90
```diff
@@ -209,41 +397,45 @@ end subroutine interior_to_halo_communication
...
+        subroutine field_mpi_alloc(ncomp)
+            integer, intent(in) :: ncomp
+            integer             :: zlen, ylen, xlen
 
             if (l_allocated) then
                 return
             endif
             l_allocated = .true.
 
+            n_comp = ncomp
```
sjboeing, line 410: Is this needed?

sjboeing (commented):
Thanks, Matthias! I just looked at the code once more, looks good to me!

src/3d/fields/field_mpi.f90
matt-frey: Yes, it is alsways true unless `l_alloc` is present and is false. Do you prefer a mandatory flag? I thought it would be more user-friendly with an optional flag.

src/3d/fields/field_mpi.f90
matt-frey: I use this for checking when filling the buffer.

src/3d/fields/field_mpi.f90
sjboeing: Both would work, I thought the code would be easier to understand and the developer is forced to think (which I usually regard as a good thing) when it is mandatory but this is a design choice.

src/3d/fields/field_mpi.f90
sjboeing: Ah, sorry, I cannot see quickly why there are 2 different variables in this subroutine.

matt-frey: `ncomp` is input, but `n_comp` is a module variable.

matt-frey: https://github.com/EPIC-model/epic/pull/470/files#diff-2afcc8f9d8f0c020f259666a058c80a2afb4364a7e35c5d8a1e964e146e01428R487

sjboeing: OK, ncomp is global.

sjboeing:
Happy for you to merge this whenever you are ready.

src/3d/fields/field_mpi.f90
matt-frey: Ok, I'll change and merge later.

matt-frey: Changed. See https://github.com/EPIC-model/epic/pull/470/commits/f9246b253586fbddbb4c3a23cc210e3753979653.

PR #472, opened 2023-05-07
Use request-based RMA

matt-frey:
Instead of `MPI_Get/MPI_Put` we use `MPI_RGet/MPI_RPut`. A MPI_R*` routine allocates a communication request object that can be use to wait for completion. See for example https://www.open-mpi.org/doc/v4.1/man3/MPI_Rget.3.php


src/3d/parcels/parcel_nearest.f90
```diff
@@ -788,6 +792,7 @@ subroutine resolve_tree(isma, iclo, rclo, n_local_small)
                     endif
                 enddo
 
+                call MPI_Waitall(n_local_small, requests, MPI_STATUSES_IGNORE, comm%err)
```
sjboeing, line 795: Do we need both a waitall and a barrier (line below)?

sjboeing: (this happens at a few places)

matt-frey: Yes, it is needed, it does not work without. See also https://stackoverflow.com/a/48372747.

sjboeing: Happy for you to bring this in after testing. Is the advantage that we explicitly look for completion?

matt-frey: Yes, we explicitly check. I was mainly worried when removing `MPI_Win_sync` that we face synchronisation issues between private and public copies of memory. I will test the nearest comparison first, until then this PR remains in draft mode and won't be merged.

matt-frey:
We have now compared almost 23.5 billion merges. The parallel version returned the same result as the serial version.

*Test 1*
>Total number of samples:       12000
>Seeds:                         [45, 89, 351, 874]
>MPI ranks:                     [16, 32, 64]
>Number of parcels per sample:  1310720
>Number of parcels per cell:    40
>Number of fails:               0
>Number of merges:              5802331176

*Test 2 with 16, 32 and 64 MPI ranks; seed 875*
>#samples, #fails, #merges:  13200 0 6382621749

*Test 3 with with 16, 32 and 64 MPI ranks; seed 43*
>#samples, #fails, #merges:  11800 0 5705727073

*Test 4 with 72, 96 and 120 MPI ranks; seeds 45 89 351 874*
>#samples, #fails, #merges:  11500 0 5560559699

PR #477, opened 2023-05-07
Fix MPI unit tests

matt-frey:
This PR
* fixes some unit tests after changing the initialisation interface of `mpi_reverse.f90` and `parcel_nearest.f90`.
* fixes the `Floating-point exception - erroneous arithmetic operation` error happening in `set_hyperbolic_functions` of `inversion_utils.f90` due to underflow. This fix is only applied in debug mode.
* slightly misaligns the position of `x` or `y` in `test_mpi_nearest_3.f90` and `test_mpi_nearest_4.f90` as round-off errors cause the parcels to be assigned to different grid points that then results in `Merge error: no near neighbour found`.
* fixes `n_total_parcels` in `test_mpi_gradient_correction_3d.f90`.

### DIFF
--- a/unit-tests/3d/test_mpi_nearest_3.f90
+++ b/unit-tests/3d/test_mpi_nearest_3.f90
@@ -100,7 +100,9 @@ program test_mpi_nearest_3
             iz = k
 
             x = lower(1) + (0.5d0 + dble(ix)) * dx(1)
-            y = lower(2) + (0.5d0 + dble(iy)) * dx(2)
+            ! misalign with centre to avoid round-off errors in
+            ! grid point assignment
+            y = lower(2) + (0.500001d0 + dble(iy)) * dx(2)
             z = lower(3) + (0.5d0 + dble(iz)) * dx(3)
 
             ! big parcel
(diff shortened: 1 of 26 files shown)


unit-tests/3d/test_mpi_nearest_3.f90
```diff
@@ -96,7 +100,9 @@ subroutine cell_placement(l, i, j, k)
             iz = k
 
             x = lower(1) + (0.5d0 + dble(ix)) * dx(1)
```
sjboeing, line 102: Do we want to do the same with x, just in case?

unit-tests/3d/test_mpi_nearest_4.f90
```diff
@@ -95,7 +99,9 @@ subroutine cell_placement(l, i, j, k)
             iy = j
             iz = k
 
-            x = lower(1) + (0.5d0 + dble(ix)) * dx(1)
+            ! misalign with centre to avoid round-off errors in
+            ! grid point assignment
+            x = lower(1) + (0.500001d0 + dble(ix)) * dx(1)
```
sjboeing, line 104: Do we want to apply this to both x and y?

unit-tests/3d/test_mpi_nearest_3.f90
matt-frey: It is not necessary in `x` for this case as we only check horizontally.

PR #479, opened 2023-05-08
No allocatable strings

matt-frey:
When compiling with `-Wuninitialized -Wmaybe-uninitialized` we get warnings of "maybe uninitialised" variables. The reason are  `allocatable` characters. We get rid of them by specifying a size.

### DIFF
--- a/analyse/2d/genspec2d.f90
+++ b/analyse/2d/genspec2d.f90
@@ -197,9 +197,9 @@ program genspec
         end subroutine read_data
 
         subroutine write_spectrum
-            logical                   :: exists = .false.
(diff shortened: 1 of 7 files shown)


matt-frey:
I have been running the code like this already on Archer2.

matt-frey:
@sjboeing We may also completely ignore this change as it is again a compiler bug. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91442 (Here is also a stackoverflow discussion https://stackoverflow.com/questions/61728557/fortran-program-using-character-variable-with-allocatable-length-always-shows-a).

PR #484, opened 2023-05-14
Fix the calculation of the vorticity correction

matt-frey:


sjboeing (approved):
Thanks Matthias, looks right to me.

PR #487, opened 2023-05-23
3rd order RK stepper

matt-frey:
This PR adds the 3rd order RK stepper. Users can select between 3rd and 4th order RK.


sjboeing (commented):
Thanks, Matthias, looks good!

PR #489, opened 2023-05-26
Use amax instead vmax as parcel splitting criterion

matt-frey:


matt-frey:
@sjboeing This does still need some testing.

unit-tests/3d/test_mpi_parcel_split.f90
```diff
@@ -58,6 +57,9 @@ program test_mpi_parcel_split
     b2 = f18 * abc
     c2 = b2
 
+    abc = dsqrt(a2)
```
sjboeing, line 60: Is this the right name, or simply used here to reuse "abc"

matt-frey: I just reused the variable. I didn't want to introduce a new one just for this.

sjboeing (approved):

PR #493, opened 2023-05-30
Resize parcel container

### LINKED ISSUE #491: Adaptive parcel container size
We currently have an input parameter to increase the parcel container size. However, the container size remains fixed during a simulation and may cause the simulation to crash if its size is exceeded. Also, simulations with a lower parcel count than estimated waste memory. We should therefore adjust/check the parcel container usage after each time step.

matt-frey:
This PR
* adds two new input options (`parcel%grow_factor` and `parcel%shrink_factor`) to adjust the parcel container.
* fixes the first estimate of `max_num_parcels` to include the number of halo cells too.

Closes #491.

### DIFF
--- a/src/3d/parcels/parcel_split.f90
+++ b/src/3d/parcels/parcel_split.f90
@@ -27,7 +27,7 @@ module parcel_split_mod
     ! current index for "running average"
     integer, protected :: ri = 0
 
-    private :: dh, n_previous_splits, ri
+    private :: dh, n_previous_splits, ri, adapt_container_size
 
     integer :: split_timer
 
@@ -37,24 +37,13 @@ module parcel_split_mod
 
     contains
 
-        ! Split large parcels (volumes larger than vmax) or
-        ! parcels with aspect ratios larger than parcel%lambda_max.
-        subroutine parcel_split
-            double precision     :: B(5)
-            double precision     :: vol, lam
-            double precision     :: D(3), V(3, 3)
-            integer              :: last_index
-            integer              :: n, n_thread_loc
-            integer              :: pid(2 * n_parcels)
-            integer, allocatable :: invalid(:)
+        ! Adapt the container size. This routine is invoked before
+        ! the actual splitting. It estimstes the final number of parcels
+        ! according to previous split calls. If the estimated number exceeds
+        ! the container size, a container resize operation is invoked. If the
+        ! estimated number is below a threshold, the container size is reduced.
+        subroutine adapt_container_size
             integer              :: n_estimate, shrunk_size, grown_size
-#ifdef ENABLE_VERBOSE
-            integer              :: orig_num
-
-            orig_num = n_total_parcels
-#endif
-
-            call start_timer(split_timer)
 
             ! Estimate final number of parcels based on previous number of splits:
             n_estimate = nint(dble(sum(n_previous_splits)) / dble(size(n_previous_splits))) &
@@ -68,6 +57,29 @@ module parcel_split_mod
             else if (n_estimate < shrunk_size) then
                 call parcel_resize(shrunk_size)
             endif
+
+        end subroutine adapt_container_size
+
+        !::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+
+        ! Split large parcels (volumes larger than vmax) or
+        ! parcels with aspect ratios larger than parcel%lambda_max.
+        subroutine parcel_split
+            double precision     :: B(5)
+            double precision     :: vol, lam
+            double precision     :: D(3), V(3, 3)
+            integer              :: last_index
+            integer              :: n, n_thread_loc
+            integer              :: pid(2 * n_parcels)
+            integer, allocatable :: invalid(:)
+#ifdef ENABLE_VERBOSE
+            integer              :: orig_num
+
+            orig_num = n_total_parcels
+#endif
+            call adapt_container_size
+
+            call start_timer(split_timer)
 
             last_index = n_parcels
 
--- a/src/3d/parcels/parcel_split.f90
+++ b/src/3d/parcels/parcel_split.f90
@@ -47,7 +47,7 @@ module parcel_split_mod
             integer              :: n, n_thread_loc
             integer              :: pid(2 * n_parcels)
             integer, allocatable :: invalid(:)
-            integer              :: n_estimate
+            integer              :: n_estimate, shrunk_size, grown_size
 #ifdef ENABLE_VERBOSE
             integer              :: orig_num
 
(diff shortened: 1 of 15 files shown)


src/3d/parcels/parcel_split.f90
```diff
@@ -16,34 +19,65 @@ module parcel_split_mod
...
+        subroutine adapt_container_size
+            integer              :: n_estimate, shrunk_size, grown_size
+
+            ! Estimate final number of parcels based on previous number of splits:
+            n_estimate = nint(dble(sum(n_previous_splits)) / dble(size(n_previous_splits))) &
+                       + n_parcels
+
+            shrunk_size = nint(parcel%shrink_factor * max_num_parcels)
+
+            if (n_estimate > max_num_parcels) then
```
sjboeing, line 54: This if statement has no buffer, I think. So if your estimate is off, you can easily go over max_num_parcels.

src/3d/parcels/parcel_split.f90
```diff
@@ -16,34 +19,65 @@ module parcel_split_mod
...
+                grown_size = nint(parcel%grow_factor * max_num_parcels)
+                call parcel_resize(grown_size)
+            else if (n_estimate < shrunk_size) then
```
sjboeing, line 57: Similarly, this means that if your estimate is just (even 1) below shrunk_size, you could leave very little buffer.

src/3d/parcels/parcel_split.f90
```diff
@@ -16,34 +19,65 @@ module parcel_split_mod
...
 
+
     contains
 
+        ! Adapt the container size. This routine is invoked before
+        ! the actual splitting. It estimstes the final number of parcels
+        ! according to previous split calls. If the estimated number exceeds
+        ! the container size, a container resize operation is invoked. If the
+        ! estimated number is below a threshold, the container size is reduced.
+        subroutine adapt_container_size
```
sjboeing, line 45: I would probably try to keep n_estimate between 0.7*max_num_parcels and 0.85*max_num_parcels. This leaves some buffer for not running out of indices with a "wrong estimate" and also some buffer so you do not need to keep changing the array size every time step.

If your estimate is above 0.85*max_num_parcels, increase the buffer so that your estimate is 0.75*new_max_num_parcels, i.e new_max_num_parcels=nint(f43*n_estimate). In that way, you increase the buffer in such a way that you are not likely to face the same problem in the next time step if the number of parcels varies slowly enough. 

Similarly, if your estimate goes below 0.7*max_num_parcels, decrease max_num_parcels so that your n_estimate=0.8*new_max_num_parcels, i.e. new_max_num_parcels=f54*n_estimate.

The numbers are intuitive guesses, but should provide a decent level of additional safety.

sjboeing: Alternatively, you could reset so that new_max_num_parcels=n_estimate/0.775 coming from each direction.

src/3d/parcels/parcel_split.f90
matt-frey: I added some buffer. See 013d28a06f085f6fb2e27f60bca86d19f46772f8.

matt-frey: Note: The buffer for growing was `grown_size`. We have now even more.

src/3d/utils/options.f90
```diff
@@ -77,12 +77,16 @@ module options
     !
     type parcel_type
         double precision :: size_factor      = 1.0d0    ! factor to increase max. number of parcels
+        double precision :: grow_factor      = 0.5d0    ! factor to increase the parcel container size
```
sjboeing, line 80: Should be bigger than 1? I would suggest using 1.2 or so, but see also other comment on growth.

src/3d/utils/options.f90
```diff
@@ -77,12 +77,16 @@ module options
...
+                                                        ! in the parcel splitting routine
+        double precision :: shrink_factor    = 0.5d0    ! factor to reduce the parcel container size
```
sjboeing, line 82: Probably the shrink factor can be something like 0.8, i.e. we can use smaller steps to shrink.

src/3d/parcels/parcel_split.f90
```diff
@@ -16,34 +19,65 @@ module parcel_split_mod
...
+            integer              :: n_estimate, shrunk_size, grown_size
+
+            ! Estimate final number of parcels based on previous number of splits:
+            n_estimate = nint(dble(sum(n_previous_splits)) / dble(size(n_previous_splits))) &
+                       + n_parcels
+
+            shrunk_size = nint(parcel%shrink_factor * max_num_parcels)
+
+            if (n_estimate > int(f34 * max_num_parcels)) then
+                grown_size = nint(parcel%grow_factor * max_num_parcels)
```
sjboeing, line 55: You could use something like:
grown_size = nint(parcel%grow_factor * f43 * n_estimate), i.e. you would use n_estimate in determining how big the container needs to be. Won't be very different for a slowly varying parcel number though.
See below for note on grow_factor.

src/3d/parcels/parcel_split.f90
```diff
@@ -23,49 +26,98 @@ module parcel_split_mod
...
+
+            n_indices = size(indices)
+
+            !------------------------------------------------------------------
+            ! Adapt container size if needed:
+
+            ! we get additional "n_indices" parcels
+            n_required = n_parcels + n_indices
+
+            shrunk_size = max(n_required, max_num_parcels)
```
sjboeing, line 88: Can we not simply use max_num_parcels here, given the check below?

sjboeing (commented):
Looks very good, Matthias! Can you test this on a case where it is active (maybe monitoring if it works, i.e. how often it is called)?

matt-frey:
Yes, I am always running the Beltrami test case to see if it works.

src/utils/armanip.f90
```diff
@@ -0,0 +1,73 @@
...
+
+            ncomp = size(attr, dim=1)
+            copy_size = size(attr, dim=2)
+
+            if (present(old_size)) then
+                copy_size = old_size
+            endif
+
+            if (new_size < copy_size) then
+                call mpi_exit_on_error("armanip::resize_array_2d: new_size < old_size")
```
sjboeing, line 62: Maybe this should read "new_size < copy_size" to prevent confusion. The array can shrink, but must remain bigger than n_parcels.

src/utils/armanip.f90
```diff
@@ -0,0 +1,73 @@
...
+            double precision, allocatable                :: buffer(:)
+            integer                                      :: copy_size
+
+            copy_size = size(attr)
+            if (present(old_size)) then
+                copy_size = old_size
+            endif
+
+            if (new_size < copy_size) then
+                call mpi_exit_on_error("armanip::resize_array_1d: new_size < old_size")
```
sjboeing, line 34: Maybe this should read "new_size < copy_size" to prevent confusion. The array can shrink, but must remain bigger than n_parcels.

sjboeing (commented):
This looks good to me. The main thing is to test the parallelism as a whole, I think.

matt-frey:
I already compared this with the version without resizing. No significant difference.

src/utils/armanip.f90
sjboeing: Ah, yes, maybe the word "old_size" is confusing me a bit, as it it actually often the number of parcels rather than the actual array size.

matt-frey: I have already changed to `copy_size`.

sjboeing: Ah, I think the issue comes from naming it "old_size" in the input arguments. In my mind, the "old_size" refers to the old size of the array, and the "copy_size" to the size of the part that is meant to be copied to the new array (so the input argument would then be copy_size, and copy_size would be set to old_size if it is absent). But it is just a matter of variable naming.

matt-frey: I exchanged the names, see https://github.com/EPIC-model/epic/pull/493/commits/f4df3cb54b75dfe47304debd417c66c029bbdebf.

PR #496, opened 2023-06-07
Communicate veta and vtau

matt-frey:
We must communicate Veta and Vtau as well. Another solution: We do not reuse them, but calculate them all the time in `vol2grid`, `par2grid` and the parcel corrections. We must communicate them due to the parcel halo swap calls.

### DIFF
--- a/src/3d/parcels/parcel_container.f90
+++ b/src/3d/parcels/parcel_container.f90
@@ -5,7 +5,11 @@
 module parcel_container
     use options, only : verbose
     use parameters, only : extent, extenti, center, lower, upper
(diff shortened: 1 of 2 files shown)


sjboeing (commented):
This looks right to me. I am now wondering though if it is really needed since I think we recalculate every time step (i.e. is the parcel position changed between when we first calculate it and when we reuse it).

matt-frey:
We mainly need this for the parcel correction as the Laplace calls a halo swap.

matt-frey:
I agree that in the RK stepper this is not needed as we recalculate them every time.

sjboeing:
Ah, yes, agreed!

PR #498, opened 2023-06-12
Use MPI_Win_allocate instead of MPI_Win_create

matt-frey:
According to https://enccs.github.io/intermediate-mpi/one-sided-sync/#passive-target-communication (accessed 12 June 2023) it is safer to use MPI_Win_allocate for passive target communication.


sjboeing (approved):
Thanks, looks good. Happy to approve if it works on a case with a substantial amount of merging.

matt-frey:
Yes, I am running the comparison between serial and parallel merging. Already 2 billion mergers.

sjboeing:
Have you run e.g. Beltrami as well? With those 2 tests you would probably cover a lot.

matt-frey:
No, I have not run Beltrami as this PR only changes the allocation of the memory and not the underlying algorithm.

PR #500, opened 2023-06-13
Pull optimizations to serial version

matt-frey:
This PR applies all code optimizations to the serial version of EPIC, these are
* option to run RK3 (see #487)
* Scherzinger eigenvalue/-vector solver (see #488 and #495)
* amax instead of vmax (see #489)
* container resizing (see #493)

### DIFF
--- a/src/3d/epic3d.f90
+++ b/src/3d/epic3d.f90
@@ -31,7 +31,6 @@ program epic3d
     use utils, only : write_last_step, setup_output_files        &
                     , setup_restart, setup_domain_and_parameters &
                     , setup_fields_and_parcels
-    use parameters, only : max_num_parcels
     use options, only : rk_order
     implicit none
 
--- a/src/3d/epic3d.f90
+++ b/src/3d/epic3d.f90
@@ -121,7 +121,7 @@ program epic3d
 
                 call merge_parcels(parcels)
 
-                call parcel_split(parcels, parcel%lambda_max)
+                call parcel_split
 
                 do cor_iter = 1, parcel%correction_iters
                     call apply_laplace((cor_iter > 1))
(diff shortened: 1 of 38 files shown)


src/3d/epic3d.f90
```diff
@@ -27,11 +27,12 @@ program epic3d
     use inversion_utils, only : init_inversion
     use parcel_interpl, only : grid2par_timer, par2grid_timer
     use parcel_init, only : init_timer
-    use ls_rk4, only : ls_rk4_alloc, ls_rk4_dealloc, ls_rk4_step, rk4_timer
+    use ls_rk, only : ls_rk4_alloc, ls_rk4_dealloc, ls_rk_step, rk_timer, ls_rk_setu
```
sjboeing, line 30: Maybe the alloc routines should at some point ust be called ls_rk_alloc etc. But this should not matter for functionality.

sjboeing (commented):
Thanks for doing this, Matthias. Do you want to set rk_order==3 for some setups or compare with/leave the code at rk_order 4?

src/3d/epic3d.f90
matt-frey: I have just remove those routines.

sjboeing (commented):
Looks good! One thing I found is that I need to remove max_vratios in order to run. Actually it would be good to make the changes that force not setting max_vratios for the parallel version as well. 

Now wondering if we may want a slightly smaller a_max to guard against changes in a within the time step to a degree. I think it is not likely to crash on this, but there may be a theoretical possibility.

matt-frey:
Ok, thanks for letting me know. I will fix the max_vratio issue.

matt-frey:
I removed `max_vratio.

sjboeing (approved):

PR #502, opened 2023-06-27
Fixes issue with zero-indexing in inversion module

sjboeing:
This small fix to the inversion (indices start with zero instead of 1) seems to solve some of the issues with reproducibility in parallel.


daviddritschel (approved):
Looks good!

matt-frey:
Well spotted!

matt-frey (approved):

PR #504, opened 2023-06-30
Avoid array temporaries

matt-frey:
Fortran creates temporary arrays when passing sub-ranges to subroutines.

### DIFF
--- a/src/3d/inversion/inversion.f90
+++ b/src/3d/inversion/inversion.f90
@@ -373,16 +373,26 @@ module inversion_mod
             call diffx(fs, ds)
             call fftxys2p(ds, f(:, :, :, I_X))
 
+            !$omp parallel workshare
+            f(  -1, :, :, I_X) = two * f( 0, :, :, I_X) - f(   1, :, :, I_X)
+            f(nz+1, :, :, I_X) = two * f(nz, :, :, I_X) - f(nz-1, :, :, I_X)
+            !$omp end parallel workshare
+
             ! calculate df2/dy
             call fftxyp2s(f(:, :, :, I_Y), fs)
             call diffy(fs, ds)
             call fftxys2p(ds, f(:, :, :, I_Y))
 
+            !$omp parallel workshare
+            f(  -1, :, :, I_Y) = two * f( 0, :, :, I_Y) - f(   1, :, :, I_Y)
+            f(nz+1, :, :, I_Y) = two * f(nz, :, :, I_Y) - f(nz-1, :, :, I_Y)
+            !$omp end parallel workshare
+
             ! calculate df3/dz
             call central_diffz(f(:, :, :, I_Z), div)
 
             ! div = df1/dx + df2/dy + df3/dz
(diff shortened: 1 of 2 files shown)


src/3d/inversion/inversion.f90
```diff
@@ -360,10 +360,11 @@ end subroutine vorticity_tendency
 
         !::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 
-        ! Note: f is overwritten
+        ! Note: f is overwritten; only the range div(0:nz, :, :) contains
+        ! valid data
         subroutine divergence(f, div)
             double precision, intent(inout) :: f(-1:nz+1, box%hlo(2):box%hhi(2), box%hlo(1):box%hhi(1), n_dim)
-            double precision, intent(out)   :: div(0:nz, box%hlo(2):box%hhi(2), box%hlo(1):box%hhi(1))
+            double precision, intent(out)   :: div(-1:nz+1, box%hlo(2):box%hhi(2), box%hlo(1):box%hhi(1))
```
sjboeing, line 367: Should there be default output for div at -1 or nz+1 (if only for safety)? I am not sure about this.

src/3d/inversion/inversion_utils.f90
```diff
@@ -482,6 +483,32 @@ subroutine central_diffz(fs, ds)
...
 
+        end subroutine central_diffz_semi_spectral
+
+        !::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+
+        !Calculates df/dz for a field f using 2nd-order differencing.
+        !Here df = df/dz. In physical space.
+        subroutine central_diffz(f, df)
+            double precision, intent(in)  :: f(-1:nz+1,  box%hlo(2):box%hhi(2), box%hlo(1):box%hhi(1))
+            double precision, intent(out) :: df(-1:nz+1, box%hlo(2):box%hhi(2), box%hlo(1):box%hhi(1))
```
sjboeing, line 494: Sam question here.

sjboeing (commented):
Thanks for writing this Matthias! I can see why you initially went for the other approach, but thought it would be worth trying to get rid of the compiler warnings. Not quite sure whether to do anything (by default) with the values of the output fields at z=-1 and z=nz+1.

src/3d/inversion/inversion.f90
matt-frey: We can use linear extrapolation.

matt-frey:
We could also get rid of these warnings by explicitly creating a temporary variable ourselves.

sjboeing:
What do you think is best practice/most efficient? I would be OK with either solution. Guess this is the next one to merge after the period BCs PR.

matt-frey:
Avoiding temporary arrays is definitely more efficient as we do not copy memory. The indices `-1` and `nz+1` and we can either set them to zero or (maybe better) use linear extrapolation.

matt-frey:
My suggested change: https://github.com/EPIC-model/epic/pull/504/commits/5759bceae46483e0074891346e54e4fa4acc5bb6

sjboeing:
Let's use the extrapolation, which I agree should be safer.

PR #505, opened 2023-07-03
Parcel interpolation simplifications

sjboeing:
Implemented this change here, it seems to work and is slightly more efficient.


matt-frey (changes requested):
I also changed it locally. However, I also removed the routine `get_weights` completely as it is not used anywhere else.

sjboeing:
Thanks @matt-frey: yes, removing the additional routine seems sensible. I am just double-checking this works.

matt-frey (approved):
Looks good.

matt-frey:
This may also be cherry-picked to the branch `3d-mpi-model`.

PR #510, opened 2023-07-14
Remove max_vratio from configuration files

matt-frey:


sjboeing (commented):
I think there is another reference to vmax_ratio in
3d/utils/options.f90
(in the parcel type)

Otherwise this looks good

matt-frey:
Indeed. Removed. I'll merge.

PR #514, opened 2023-07-19
Simplify parcel merge routine call

matt-frey:
Remove subroutine input argument. We do not need to pass the parcel container.


daviddritschel (approved):
Looks like mainly reformatting declarations

matt-frey:
Indeed. The way we call merging is now similar to splitting. Cleaner.

PR #520, opened 2023-07-24
Write specific fields only

matt-frey:
This PR
* changes the writing of fields: Fields can be enabled and disabled.
* reduces the amount of information printed in verbose mode

### DIFF
--- a/src/3d/parcels/parcel_container.f90
+++ b/src/3d/parcels/parcel_container.f90
@@ -231,7 +231,7 @@ module parcel_container
         subroutine parcel_replace(n, m)
             integer, intent(in) :: n, m
 
-#ifdef ENABLE_VERBOSE
+#if defined (ENABLE_VERBOSE) && !defined (NDEBUG)
             if (verbose) then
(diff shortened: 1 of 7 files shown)


src/3d/parcels/parcel_container.f90
```diff
@@ -231,7 +231,7 @@ end function get_dely
         subroutine parcel_replace(n, m)
             integer, intent(in) :: n, m
 
-#ifdef ENABLE_VERBOSE
+#if defined (ENABLE_VERBOSE) && !defined (NDEBUG)
```
sjboeing, line 234: Can you let me know why this combination is needed now?

matt-frey: I just wanted to reduce the amount of info printed to the standard output in verbose mode. This output is mainly useful for debugging.

sjboeing (approved):
I just had a quick look, it looks good. I just left one comment. Happy for you to merge when tested on some cases (e.g. beltrami and moist, that should cover most situations). Left one comment where I just wasn't quite sure of the logic.

matt-frey:
I have already tested for Beltrami. All good.

PR #522, opened 2023-07-27
Update Python packages

matt-frey:


sjboeing (approved):
Good idea to keep these up-to-date.

PR #527, opened 2023-08-11
Fix the write frequency

matt-frey:
This PR changes / fixes the frequency of writes.


sjboeing (approved):
Looks good, Matthias!

PR #530, opened 2023-08-18
Increase buffer in parcel communication

matt-frey:
The parcel communication buffers do not include the vertical halo cells. We must increase their size.


matt-frey:
There is a new issue to improve the current measure, see #533.

PR #536, opened 2023-11-20
Overwrite default Zenodo metadata

matt-frey:
Add the ORCID and fix the names in the metadata. See also
* https://developers.zenodo.org/#update-schedule
* https://developers.zenodo.org/#representation


sjboeing (approved):
This looks good to me. Not sure if the umlaut can be a problem.

PR #537, opened 2023-11-23
Add missing field statistics timer stop

matt-frey:
Only the root rank writes the field statistics. All other MPI ranks leave early, hence they must stop their timers.


sjboeing (approved):
Thanks, Matthias. I didn't know the ordering in spectral space by heart but this looks consistent with the filter.

PR #538, opened 2023-11-23
Calculate squared buoyancy frequency only if not provided

matt-frey:
We currently do not calculate the (squared) buoyancy frequency in restart mode. This PR changes to the buoyancy frequency (as input) and calculates only if not provided.


sjboeing (approved):
Thanks, Matthias!

PR #540, opened 2023-11-26
Code to fix issue with netcdf synchoronisation on ARCHER2 for some

sjboeing:
This puts in additional barriers when reading fields from netCDF. I am not exactly sure why they are needed, but they solve some issues I have been experiencing.


matt-frey:
This PR is not necessary anymore. Issue seems to be resolved.

PR #543, opened 2023-11-29
Selective parcel attribute writes

matt-frey:
Allow specifying which parcel attributes to write to a file.


sjboeing (approved):
Looks good from a quick inspection.

PR #544, opened 2023-11-29
Collect merger statistics

matt-frey:
Write parcel merger statistics to the netCDF file.

### DIFF
--- a/src/3d/parcels/parcel_merge.f90
+++ b/src/3d/parcels/parcel_merge.f90
@@ -29,8 +29,13 @@ module parcel_merging
     ! number of parcel merges (is reset in every write step)
     integer :: n_parcel_merges = 0
 
-    private :: geometric_merge, &
-               do_group_merge
+    ! number of merging parcels (up to 10 supported, all others are put into index 10)
+    ! note that array index 1 corresponds to 2-way merging
+    integer :: n_way_parcel_mergers(10) = 0
+
+    private :: geometric_merge,     &
+               do_group_merge,      &
+               collect_merge_stats
 
     contains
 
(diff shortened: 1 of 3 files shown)


src/3d/parcels/parcel_merge.f90
```diff
@@ -318,4 +332,42 @@ subroutine geometric_merge(isma, iclo, n_merge)
...
+            unique = 0
+            number = 0 ! note that "number" does not include the 'iclo' parcel in the count
+
+            !------------------------------------------------------------------
+            ! Find unique 'iclo' indices and the total number of parcels
+            ! merging with them:
+            last = 0
+            do m = 1, n_merge
+                ic = iclo(m)
+                n_big_close = n_big_close + merge(1, 0, parcels%volume(ic) > vmin)
```
sjboeing, line 353: Does this really save over an if statement?

sjboeing: Happy to keep this, just wondering if the compiler wouldn't take care of this.

src/3d/parcels/parcel_merge.f90
```diff
@@ -318,4 +332,42 @@ subroutine geometric_merge(isma, iclo, n_merge)
...
+                j = findloc(unique(1:last), value = ic, dim=1)
```
sjboeing, line 354: Wondering if the computational cost here is unpredictable or could be high for many mergers as a findloc needs to be done for every merge. I guess the alternative would be to keep track in an array which may be oversized for what it really does?

sjboeing (commented):
Just left some comments. Looks good though I was wondering if we can avoid findloc.

sjboeing (approved):
Happy for you to merge, Matthias. I was wondering (see slack) if there is an argument for using a temporary integer array here, rather than reusing an array, if we use temporary arrays elsewhere (and this is not the point in the code where the memory footprint is largest).

src/3d/parcels/parcel_merge.f90
matt-frey: Changed.

PR #549, opened 2024-01-09
Damping implementation

sjboeing:
Work in progress

### DIFF
--- a/examples/moist.config
+++ b/examples/moist.config
@@ -9,7 +9,7 @@
  ! output info
  !
  output%field_freq         = 50             ![s] write after these many seconds to the field NetCDF file
- output%parcel_freq        = 50             ![s] write after these many seconds to the parcel NetCDF file
+ output%parcel_freq        = 200             ![s] write after these many seconds to the parcel NetCDF file
  output%parcel_stats_freq  = 50             ![s] write after these many seconds to parcel stats NetCDF file
  output%field_stats_freq   = 50             ![s] write after these many seconds to the field stats NetCDF file
  output%write_fields       = .true.         ! enable / disable field dump
--- a/examples/moist.config
+++ b/examples/moist.config
@@ -40,5 +40,5 @@
  ! damping info
  !
  damping%l_vorticity     = .false.
- damping%l_scalars       = .true.
+ damping%l_scalars       = .false.
 /
--- a/examples/moist.config
+++ b/examples/moist.config
@@ -35,4 +35,10 @@
  time%limit              = 1000.0           ! time limit (s)
  time%alpha              = 0.2              ! scaling factor for the strain and buoyancy gradient time step
  time%precise_stop       = .false.          ! time limit exact
+
+ !
+ ! damping info
+ !
+ damping%l_vorticity     = .false.
+ damping%l_scalars       = .true.
 /
--- a/src/3d/parcels/parcel_damping.f90
+++ b/src/3d/parcels/parcel_damping.f90
@@ -35,22 +35,16 @@ module parcel_damping
 
             if (damping%l_vorticity .or. damping%l_scalars) then 
                 call par2grid(.false.)
-                call perturbation_damping(dt, damping%l_vorticity, damping%l_scalars, &
-                     damping%vorticity_prefactor, damping%scalars_prefactor, .true.)
+                call perturbation_damping(dt, .true.)
             end if
 
         end subroutine parcel_damp
 
         ! 
         ! @pre 
-        subroutine perturbation_damping(dt, l_vorticity, l_scalars, &
-                                        vorticity_prefactor, scalars_prefactor, l_reuse)
+        subroutine perturbation_damping(dt, l_reuse)
             double precision, intent(in)  :: dt
-            logical, intent(in)   :: l_vorticity
-            logical, intent(in)   :: l_scalars
-            logical   :: l_reuse
-            double precision, intent(in)   :: vorticity_prefactor
-            double precision, intent(in)   :: scalars_prefactor
+            logical, intent(in)   :: l_reuse
             integer                       :: n, p, l 
             double precision              :: points(3, n_points_p2g)
             double precision              :: pvol
@@ -89,16 +83,16 @@ module parcel_damping
                 do p = 1, n_points_p2g
                     call trilinear(points(:, p), is, js, ks, weights)
                     weight = point_weight_p2g * weights
-                    if (l_vorticity) then
+                    if (damping%l_vorticity) then
                         ! Note this exponential factor can be different for vorticity/scalars
-                        time_fact = one - exp(-vorticity_prefactor * strain_mag(ks:ks+1, js:js+1, is:is+1) * dt)
+                        time_fact = one - exp(-damping%vorticity_prefactor * strain_mag(ks:ks+1, js:js+1, is:is+1) * dt)
                         do l = 1,3
                             vortend(l) = vortend(l)+sum(weight * time_fact * (vortg(ks:ks+1, js:js+1, is:is+1, l) &
                                        - parcels%vorticity(l,n)))
                         enddo
                     endif
-                    if (l_scalars) then
-                        time_fact = one - exp(-scalars_prefactor * strain_mag(ks:ks+1, js:js+1, is:is+1) * dt)
+                    if (damping%l_scalars) then
+                        time_fact = one - exp(-damping%scalars_prefactor * strain_mag(ks:ks+1, js:js+1, is:is+1) * dt)
 #ifndef ENABLE_DRY_MODE
                         humtend = humtend + sum(weight * time_fact * (humg(ks:ks+1, js:js+1, is:is+1) - parcels%humidity(n)))
                         buoytend = buoytend + sum(weight * time_fact * (dbuoyg(ks:ks+1, js:js+1, is:is+1) - parcels%buoyancy(n)))
@@ -108,12 +102,12 @@ module parcel_damping
                     endif
                 enddo
                 ! Add all the tendencies only at the end
-                if (l_vorticity) then
+                if (damping%l_vorticity) then
                     do l=1,3
                         parcels%vorticity(l,n) = parcels%vorticity(l,n) + vortend(l)
                     enddo
                 endif
-                if (l_scalars) then
+                if (damping%l_scalars) then
 #ifndef ENABLE_DRY_MODE
                     parcels%humidity(n) = parcels%humidity(n) + humtend
 #endif
--- a/src/3d/parcels/parcel_damping.f90
+++ b/src/3d/parcels/parcel_damping.f90
@@ -51,7 +51,7 @@ module parcel_damping
             logical   :: l_reuse
             double precision, intent(in)   :: vorticity_prefactor
             double precision, intent(in)   :: scalars_prefactor
-            integer                       :: n, p, l, ii, jj, kk
+            integer                       :: n, p, l 
             double precision              :: points(3, n_points_p2g)
             double precision              :: pvol
             ! tendencies need to be summed up between associated 4 points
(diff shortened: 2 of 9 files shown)


src/3d/utils/options.f90
```diff
@@ -189,6 +200,11 @@ subroutine write_netcdf_options(ncid)
             call write_netcdf_attribute(ncid, "initial", time%initial)
             call write_netcdf_attribute(ncid, "precise_stop", time%precise_stop)
             call write_netcdf_attribute(ncid, "alpha", time%alpha)
+
```
sjboeing, line 203: Added the prefix damping here. Is this OK?

examples/moist.config
```diff
@@ -9,7 +9,7 @@
  ! output info
  !
  output%field_freq         = 50             ![s] write after these many seconds to the field NetCDF file
- output%parcel_freq        = 50             ![s] write after these many seconds to the parcel NetCDF file
+ output%parcel_freq        = 200             ![s] write after these many seconds to the parcel NetCDF file
```
sjboeing, line 12: May revert this in PR, although in practice this is the setting I tend to use in order to limit parcel output.

matt-frey: I usually just turn off parcel dumping completely. `output%write_parcels = .false.`

src/3d/utils/options.f90
matt-frey: Yes, that is fine. Otherwise, we can also create a subgroup.

src/3d/parcels/parcel_damping.f90
```diff
@@ -0,0 +1,146 @@
...
+                    if(l_vorticity) then
+                      do ii=0,1
+                        do jj=0,1
+                          do kk=0,1
+                            ! Note this exponential factor can be different for vorticity/scalars
+                            time_fact(kk,jj,ii)=1.0-exp(-vorticity_prefactor*&
+                                                strain_mag(ks+kk, js+jj, is+ii)*dt)
+                          enddo
+                        enddo
+                      enddo
```
matt-frey, line 101: Is this loop really needed? Can't you just write `time_fact = 1.0- dexp(-vorticity_prefactor * strain_mag(ks:ks+1 js:js+1, is:is+1)*dt)`?

src/3d/stepper/rk_utils.f90
```diff
@@ -219,4 +219,70 @@ function get_time_step(t) result(dt)
...
+                        ! S33 = dw/dz = - (du/dx + dv/dy)
+                        strain(1, 1) = velgradg(iz, iy, ix, I_DUDX)                                   ! S11
+                        strain(1, 2) = velgradg(iz, iy, ix, I_DUDY) + f12 * vortg(iz, iy, ix, I_Z)    ! S12
+                        strain(1, 3) = velgradg(iz, iy, ix, I_DWDX) + f12 * vortg(iz, iy, ix, I_Y)    ! S13
+                        strain(2, 1) = strain(1, 2)
+                        strain(2, 2) = velgradg(iz, iy, ix, I_DVDY)                                   ! S22
+                        strain(2, 3) = velgradg(iz, iy, ix, I_DWDY) - f12 * vortg(iz, iy, ix, I_X)    ! S23
+                        strain(3, 1) = strain(1, 3)
+                        strain(3, 2) = strain(2, 3)
+                        strain(3, 3) = -(velgradg(iz, iy, ix, I_DUDX) + velgradg(iz, iy, ix, I_DVDY)) ! S33
```
matt-frey, line 266: This is the same code which we use for the time step estimate. We should put this into a function instead of duplicating the code.

src/3d/parcels/parcel_damping.f90
matt-frey: Can you maybe also add an explanation to the module how this is done? I.e. the reasoning behind these equations.

matt-frey (changes requested):
Looks generally good. However, I suggest some changes. Also, could you change to 4 white space characters instead of 2 as we use 4 throughout EPIC? Thanks.

src/3d/stepper/rk_utils.f90
sjboeing: Largely, yes. Though the output into an array is different. May try to pack it into a function.

src/3d/parcels/parcel_damping.f90
sjboeing: I think initially this did not work for me, but maybe something changed in the meantime.

sjboeing: I have revised the formulation a couple of times. A draft write-up is section 1 here: https://www.overleaf.com/project/6381d70192a0f271ee433739

src/3d/parcels/parcel_damping.f90
```diff
@@ -0,0 +1,131 @@
...
+                call par2grid(.false.)
+                call perturbation_damping(dt,damping%l_vorticity,damping%l_scalars,&
+                     damping%vorticity_prefactor,damping%scalars_prefactor,.true.)
+            end if
+
+        end subroutine parcel_damp
+
+        ! 
+        ! @pre 
+        subroutine perturbation_damping(dt,l_vorticity,l_scalars,&
```
matt-frey, line 46: Since the information of the `damping` type is globally known, you do not need to pass these.

src/3d/parcels/parcel_damping.f90
```diff
@@ -0,0 +1,131 @@
...
+            !$omp do private(n, p, l, ii, jj, kk, points, pvol, weight) &
+#ifndef ENABLE_DRY_MODE
+            !$omp& private(is, js, ks, weights, vortend, buoytend, humtend, time_fact) 
+#else
+            !$omp& private(is, js, ks, weights, vortend, buoytend, time_fact) 
+#endif
+            do n = 1, n_parcels
+                pvol = parcels%volume(n)
+#ifndef ENABLE_P2G_1POINT
+                points = get_ellipsoid_points(parcels%position(:, n), &
```
matt-frey, line 77: Is it safe to have `l_reuse = .true.`?

src/3d/parcels/parcel_damping.f90
```diff
@@ -0,0 +1,131 @@
...
+                vortend=zero
+                buoytend=zero
+#ifndef ENABLE_DRY_MODE
+                humtend=zero
+#endif
+
+                ! we have 4 points per ellipsoid
+                do p = 1, n_points_p2g
+                    call trilinear(points(:, p), is, js, ks, weights)
+                    weight = point_weight_p2g * weights
```
matt-frey, line 91: Why not assigning to `weights` so you do not need an additional variable `weight`?

matt-frey (changes requested):
Looks much better now! Thanks! I still have some suggestions. Can you maybe also use spaces between opertions, like a = b + c instead of a=b+c. It helps to improve readability. We usually also add spaces for if-statements, i.e. "if (...)" instead of "if(...)".

src/3d/parcels/parcel_damping.f90
matt-frey: It may still be good to have at least some explanatory words in the module description. Of course, you do not need to derive the theory there.

src/3d/parcels/parcel_damping.f90
sjboeing: Mainly to alert people these are no longer the weights as they came out of trilinear. Let me know if this is OK.

sjboeing:
@matt-frey: Let me know if the layout etc. looks OK now.

src/3d/parcels/parcel_damping.f90
matt-frey: Thanks for adding more info to the module description.

src/3d/stepper/rk_utils.f90
```diff
@@ -219,4 +231,35 @@ function get_time_step(t) result(dt)
...
+        ! @returns the strain magnitude for use in damping routine
+        subroutine get_strain_magnitude_field
+            double precision             :: strain(n_dim, n_dim)
+            integer                      :: ix, iy, iz
+
+            do ix = box%lo(1), box%hi(1)
+                do iy = box%lo(2), box%hi(2)
+                    do iz = 0, nz
+                        strain = get_strain(velgradg(iz, iy, ix,:), vortg(iz, iy, ix, :))
+                        strain_mag(iz, iy, ix) = sqrt(two * strain(1, 1) * strain(1, 1) +&
```
matt-frey, line 243: Why is there a two?

src/3d/stepper/rk_utils.f90
```diff
@@ -76,6 +76,51 @@ function get_dBdt(Bin, S, vorticity, volume) result(Bout)
...
+            ! S13 = 1/2 * (du/dz + dw/dx) = 1/2 * (\eta + 2 * dw/dx) = 1/2 * \eta + dw/dx
+            ! S22 = dv/dy
+            ! S23 = 1/2 * (dv/dz + dw/dy) = 1/2 * (2 * dw/dy - \xi) = dw/dy - 1/2 * \xi
+            ! S33 = dw/dz = - (du/dx + dv/dy)
+
+            strain(1, 1) = velgradgp(I_DUDX)                       ! S11
+            strain(1, 2) = velgradgp(I_DUDY) + f12 * vortgp(I_Z)    ! S12
+            strain(1, 3) = velgradgp(I_DWDX) + f12 * vortgp(I_Y)    ! S13
+            strain(2, 1) = strain(1, 2)
+            strain(2, 2) = velgradgp(I_DVDY)                                   ! S22
```
matt-frey, line 117: You may do some more alignment like here the comment `!S22` is not aligned with the others. Also `!S11`.

src/3d/parcels/parcel_damping.f90
```diff
@@ -0,0 +1,151 @@
...
+                call perturbation_damping(dt, .true.)
+            end if
+
+        end subroutine parcel_damp
+
+        ! 
+        ! @pre 
+        subroutine perturbation_damping(dt, l_reuse)
+            double precision, intent(in)  :: dt
+            logical, intent(in)   :: l_reuse
```
matt-frey, line 73: Alignment.

src/3d/stepper/rk_utils.f90
```diff
@@ -219,4 +231,35 @@ function get_time_step(t) result(dt)
...
+                        strain = get_strain(velgradg(iz, iy, ix,:), vortg(iz, iy, ix, :))
+                        strain_mag(iz, iy, ix) = sqrt(two * strain(1, 1) * strain(1, 1) +&
+                                                        strain(1, 2) * strain(1, 2) +&
+                                                        strain(1, 3) * strain(1, 3) +&
+                                                        strain(2, 1) * strain(2, 1) +&
+                                                        strain(2, 2) * strain(2, 2) +&
+                                                        strain(2, 3) * strain(2, 3) +&
+                                                        strain(3, 1) * strain(3, 1) +&
+                                                        strain(3, 2) * strain(3, 2) +&
+                                                        strain(3, 3) * strain(3, 3))
```
matt-frey, line 251: Align with `strain(1, 1)`.

matt-frey (commented):
It looks generally good now. Thanks for the hard work. There are a few places where you could do some code alignment though. I also have a question regarding the calculation of the strain magnitude.

src/3d/stepper/rk_utils.f90
sjboeing: This is the way the magnitude is defined for Smagorinsky-Lilly, e.g.:
https://en.wikipedia.org/wiki/Large_eddy_simulation#Smagorinsky%E2%80%93Lilly_model

matt-frey: But shouldn't other components have this prefactor as well?

matt-frey: Thanks for fixing.

matt-frey (approved):
Looks good to me now.

PR #553, opened 2024-02-14
Gridded strain

sjboeing:
This branch uses only gridded strain field to deform parcels. Changes to save B33 and use matrix exponentiation not yet included.

### DIFF
--- a/src/3d/fields/fields.f90
+++ b/src/3d/fields/fields.f90
@@ -50,12 +50,15 @@ module fields
         nparg,     &   ! number of parcels per grid box
         nsparg         ! number of small parcels per grid box
 
-    ! velocity strain indices
+    ! velocity strain indices (note that dw/dz is found from continuity)
     integer, parameter :: I_DUDX = 1 & ! index for du/dx strain component
                         , I_DUDY = 2 & ! index for du/dy strain component
-                        , I_DVDY = 3 & ! index for dv/dy strain component
-                        , I_DWDX = 4 & ! index for dw/dx strain component
-                        , I_DWDY = 5   ! index for dw/dy strain component
+                        , I_DUDZ = 3 & ! index for du/dz strain component
+                        , I_DVDX = 4 & ! index for dv/dy strain component
+                        , I_DVDY = 5 & ! index for dv/dy strain component
+                        , I_DVDZ = 6 & ! index for dv/dy strain component
+                        , I_DWDX = 7 & ! index for dw/dx strain component
+                        , I_DWDY = 8   ! index for dw/dy strain component
 
     contains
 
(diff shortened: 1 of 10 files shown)


src/3d/fields/fields.f90
```diff
@@ -50,12 +50,15 @@ module fields
...
 
-    ! velocity strain indices
+    ! velocity strain indices (note that dw/dz is found from continuity)
     integer, parameter :: I_DUDX = 1 & ! index for du/dx strain component
                         , I_DUDY = 2 & ! index for du/dy strain component
-                        , I_DVDY = 3 & ! index for dv/dy strain component
-                        , I_DWDX = 4 & ! index for dw/dx strain component
-                        , I_DWDY = 5   ! index for dw/dy strain component
+                        , I_DUDZ = 3 & ! index for du/dz strain component
+                        , I_DVDX = 4 & ! index for dv/dy strain component
```
matt-frey, line 57: Comment should be `dv/dx`.

src/3d/fields/fields.f90
```diff
@@ -50,12 +50,15 @@ module fields
...
+                        , I_DVDY = 5 & ! index for dv/dy strain component
+                        , I_DVDZ = 6 & ! index for dv/dy strain component
```
matt-frey, line 59: Comment should be `dv/dz`.

matt-frey (changes requested):

matt-frey (commented):
Have you tested this version? (e.g. Beltrami)

sjboeing:
I have so far run Beltrami, unit tests, and the moist bubble. Just posted some Beltrami results in the slack.

src/3d/inversion/inversion.f90
```diff
@@ -307,6 +307,13 @@ subroutine vel2vgrad(svel)
 
             call field_halo_fill_vector(velgradg, l_alloc=.true.)
 
+            !$omp parallel workshare
```
sjboeing, line 310: @matt-frey: should the halo fill be later? Should it be a halo swap instead?

sjboeing: Ah, it should be a fill but placement may be important!

matt-frey: It should be a halo fill. Halo swap is actually an accumulation. It is a bit of a misnomer.

matt-frey: As you use the vorticity there, it must have the halo being filled properly. I would check the code if this is the case.

matt-frey: It should actually not matter if you fill the halo of `velgradg` before or after -- as it overwrites data; but since you see a difference, I could imagine that the vorticity does not have the halo properly filled.

matt-frey: Here we alter the vorticity array, but without the halo being updated as it is not needed. https://github.com/EPIC-model/epic/blob/b3fadd22c09e2a583cbdfa5dd49c89420bf58871/src/3d/inversion/inversion.f90#L92C22-L92C44

matt-frey (approved):
Looks good to me.

@matt-frey This must be implemented in the surface parcel model branch as well.

PR #555, opened 2024-02-18
 Fix loop for nearest parcel check

sjboeing:
Only check local parcels after globally closest parcel determined.


matt-frey (approved):
Looks good. Thanks.

Comment on the fix: We only need to check up to `n_local_small` because after the call to `find_closest_parcel_globally` all small parcels know their closest neighbour on the original MPI rank.

PR #563, opened 2024-04-19
Add optimised rejection method

matt-frey:
This PR adds an optimised rejection method for reading a parcel dataset. If the dataset contains the MPI decomposition boxes and start indices, they will be used to selectively read parcels, i.e. not all MPI ranks read all parcels. This increases the performance for restarts from parcel files using different number of MPI ranks as the written restart file.


sjboeing (approved):
This looks right to me, Matthias. Did you compile in debug mode as well just to check if throws any warnings?

matt-frey:
Do you mean compile warnings? No, I did not get any warnings. It compiles fine in debug mode.

PR #565, opened 2024-07-06
Add labels to code

sjboeing:
Hi @matt-frey. Here is the branch that adds labels for the workshop. It is disabled by default. We may still change the dilution at some point to deal with parcels with the same labels merging, but this is not as trivial as I thought. The other thing to consider is to reset the labels every time diagnostics are created. If we do this, we can create animations of parcel trajectories (won't have time to do this before the workshop, it also means tracking over longer time scales becomes more involved).

### DIFF
--- a/src/3d/parcels/parcel_netcdf.f90
+++ b/src/3d/parcels/parcel_netcdf.f90
@@ -777,7 +777,7 @@ module parcel_netcdf
                                           dtype=NF90_INT64)
 
             nc_dset(NC_DILUTION) = netcdf_info(name='dilution',                    &
-                                          long_name='parcel dilution',             &
+                                          long_name='parcel log dilution',         &
                                           std_name='',                             &
                                           unit='1',                                &
                                           dtype=NF90_DOUBLE)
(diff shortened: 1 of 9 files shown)


src/3d/parcels/parcel_netcdf.f90
```diff
@@ -729,6 +768,19 @@ subroutine set_netcdf_parcel_info
...
+#endif
+#ifdef ENABLE_LABELS
+            nc_dset(NC_LABEL) = netcdf_info(name='label',                          &
+                                          long_name='parcel label',                &
+                                          std_name='',                             &
+                                          unit='1',                                &
+                                          dtype=NF90_INT64)
+
+            nc_dset(NC_DILUTION) = netcdf_info(name='dilution',                    &
+                                          long_name='parcel dilution',             &
```
matt-frey, line 780: What about writing `parcel log dilution`?

matt-frey (commented):
Looks generally good. Just one comment.

matt-frey (approved):

PR #566, opened 2024-07-16
Reset labels and dilution when writing to netCDF

sjboeing:
This branch resets the labels and dilution every time we write to NetCDF. Using an iterative approach, we can now map between 2 arbitrary times.


matt-frey (approved):
Looks good to me.

PR #569, opened 2024-07-24
Bugfixes to labelling

sjboeing:
Resets the labels on restarts, and fixes a wrong ifdef.

### DIFF
--- a/src/3d/parcels/parcel_container.f90
+++ b/src/3d/parcels/parcel_container.f90
@@ -434,7 +434,7 @@ module parcel_container
 #ifndef ENABLE_DRY_MODE
             parcels%humidity(n)     = buffer(IDX_HUM)
 #endif
-#ifndef ENABLE_DRY_MODE
+#ifdef ENABLE_LABELS
             parcels%label(n)        = nint(buffer(IDX_LABEL))
             parcels%dilution(n)     = buffer(IDX_DILUTION)
 #endif
--- a/src/3d/parcels/parcel_netcdf.f90
+++ b/src/3d/parcels/parcel_netcdf.f90
@@ -552,7 +552,7 @@ module parcel_netcdf
             integer, intent(in) :: first, last, pfirst
             logical             :: l_valid = .false.
             integer             :: cnt(2), start(2)
-            integer             :: num, plast
+            integer             :: num, plast, n
 
             num = last - first + 1
             plast = pfirst + num - 1
@@ -666,14 +666,16 @@ module parcel_netcdf
 #ifdef ENABLE_LABELS
             if (has_dataset(ncid, 'label')) then
                 l_valid = .true.
-                call read_netcdf_dataset(ncid, 'label', &
-                                         parcels%label(pfirst:plast), start, cnt)
             endif
             if (has_dataset(ncid, 'dilution')) then
                 l_valid = .true.
-                call read_netcdf_dataset(ncid, 'dilution', &
-                                         parcels%dilution(pfirst:plast), start, cnt)
             endif
+            ! reset the labels to Fortran index which corresponds to current label
+            ! reset the dilution to get this from time step to time step
+            do n = pfirst, plast
+               parcels%label(n) = first + n - pfirst 
+               parcels%dilution(n) = 0
+            end do
 #endif
 
 
--- a/src/3d/parcels/parcel_netcdf.f90
+++ b/src/3d/parcels/parcel_netcdf.f90
@@ -642,6 +642,7 @@ module parcel_netcdf
             endif
 
             if (has_dataset(ncid, 'y_vorticity')) then
+                l_valid = .true.
                 call read_netcdf_dataset(ncid, 'y_vorticity', &
                                          parcels%vorticity(2, pfirst:plast), start, cnt)
             endif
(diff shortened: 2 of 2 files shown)


src/3d/parcels/parcel_container.f90
```diff
@@ -434,7 +434,7 @@ subroutine parcel_deserialize(n, buffer)
 #ifndef ENABLE_DRY_MODE
             parcels%humidity(n)     = buffer(IDX_HUM)
 #endif
-#ifndef ENABLE_DRY_MODE
+#ifdef ENABLE_LABELS
```
matt-frey, line 437: Note, this bug is already fixed in the main branch with https://github.com/EPIC-model/epic/commit/6145d0190b288634a0c206d759b736f6a34cc671.

src/3d/parcels/parcel_netcdf.f90
```diff
@@ -666,14 +666,16 @@ subroutine read_chunk(first, last, pfirst)
 #ifdef ENABLE_LABELS
             if (has_dataset(ncid, 'label')) then
```
matt-frey, line 667: Note that you can just write `l_valid = (l_valid .and. has_dataset(ncid, 'label')). However, you initialise default values. Therefore, this check is not necessary at all.

matt-frey (changes requested):

src/3d/parcels/parcel_container.f90
sjboeing: Ah, thanks, don't know why my main wasn't up-to-date yet.

src/3d/parcels/parcel_netcdf.f90
sjboeing: It is a bit confusing. We do need the old values to be present for successful tracking, even though we overwrite them immediately after we write to file. One advantage of the current syntax is that it is the same as other variables, which makes it easy to read. I suggest adding a note that "existing" labels are needed for tracking. What do you think?

sjboeing: Not sure if I can think of circumstances where we have just labels and/or dilution but thought there might be a future or test case use.

matt-frey: What do you mean with "We do need the old values to be present for successful tracking, even though we overwrite them immediately after we write to file."? Here, you read parcel attributes. You do not write them. This is used in restarts.

sjboeing: What we so in a normal (non-restart) run is reinitialise the labels after writing them to disk. The reinitialisation (in the lines below) is important for backward tracking so that we can uniquely identify the "parent" parcel in the previous time-step (and then go backward by iterating this).

However, this iterative tracking relies on dumping the labels in every parcel file. If you want tracking to behave in the same way in a restart as in an uninterrupted run (at least statistically, because of the use of random numbers in merging which is currently not reproducible though we could work on this), you need to ensure that a "label dump" is present in the restart file.

matt-frey: Why then not reinitialise the labels etc only when they are present in a previous parcel dump?

sjboeing: That may be an option as well, though there may be cases where we want to start from a python-created parcel dump and add labels?

matt-frey: But even when they are not present in a previous file, you initialise them here. There is no difference. Therefore, the check if they are there or not is unnecessary.

matt-frey: To understand: Do you want the option to either read in with `read_netcdf_dataset` or reset them instead of reading?

sjboeing: What the code does is to say that a parcel dump with no vorticity/buoyancy/humidity but with labels or dilution is still a valid start dump rather than an invalid one. I am not sure about use cases but figured this may be useful for testing purposes. On the other hand, if you have no labels/dilution in that case I don't see what you would be up to.

By the way, I think we also need to add l_valid for y_vorticity?

matt-frey: I do not see a point of restarting without vorticity and buoyancy.

matt-frey: Note that we call `parcel_default` before entering this routine. You can set the labels and dilution there and then do not read in labels and dilution in the read routine at all.

matt-frey: Yes, `y_vorticity` misses the logical check `l_valid`.

sjboeing: Sure, I can remove the l_valid checks for now. We can always put it back when needed. Do you mind if I leave the rest as is and get on to other things?

matt-frey: Yes. The rest looks fine.

sjboeing: Thanks! It should be ready now.

matt-frey (approved):

matt-frey:
Do we still need this branch?

PR #572, opened 2024-07-26
Implement fixed time step

sjboeing:


src/3d/stepper/rk_utils.f90
```diff
@@ -185,8 +185,12 @@ function get_time_step(t) result(dt)
 
             gmax = local_max(1)
             bmax = local_max(2)
-
-            dt = min(time%alpha / gmax, time%alpha / bmax)
+            
+            if (time%l_use_fixed_dt) then
+                dt = time%fixed_dt
```
matt-frey, line 190: I would put this further up in the function and do an early exit. This saves some computation which is anyway discarded in this mode.

matt-frey (changes requested):

src/3d/stepper/rk_utils.f90
sjboeing: I was thinking that having the statistics on gmax and bmax could still be useful (for example, in verbose mode), especially since we aim to only use fixed_dt for testing purposes.

matt-frey: OK. However, in non-verbose mode nothing is printed. I therefore suggest to do an early extit in non-verbose mode.

sjboeing: Should be fixed now. Have a look to see if you like this (it adds a bit more code).

matt-frey (approved):
Yes, that is expected. Thanks.

PR #579, opened 2024-08-05
Matrix exp pr

sjboeing:
Just creating this PR so we can see whether to go ahead with it and you can test. I haven't been able to optimise this further. The 2D version currently still uses 2 stored elements, but would be open to making it more consistent with the 3D approach.


sjboeing:
@matt-frey: I updated the variable names.

PR #585, opened 2024-10-07
Update manual

matt-frey:
Add some more info regarding the output and model to the manual.


sjboeing:
Looks good, Matthias. Had a quick look, the only thing I noticed at the moment is that at the moment the default value of n_per_cell is set to 9. This is correct for 2D, for 3D it is 8. Don't think you actually changed this though.

matt-frey:
@sjboeing I did not change that table before. I have now. See https://github.com/EPIC-model/epic/pull/585/commits/a533fb64013bea813555da92c64d09d95ff6fb30.

PR #586, opened 2024-10-12
Fix in merging

sjboeing:
I have removed spurious periodicity checks here to ensure setups work independent of origin.


matt-frey (approved):
Looks good to me. Does this now really fix the issue you came across?

sjboeing:
Yes, this fixes the problem we had in the workshop. The only thing still needed is to run the mpi tests . Is it quick for you to run these? If not I will have a look at how to run them later this week.

matt-frey:
In principle you can only `mpirun` them.

PR #590, opened 2024-10-24
Explicit setter for type netcdf_info

matt-frey:
We are facing the issue of character string initialisation with (presumably) some compilers. The problem arises in the initialisation of the `netcdf_info` type. The assignment of character strings of shorter length than the actual type causes issues; e.g. a name like `x_velocity` suddently reads `x_volocity`. An explicit setter function fixes the issue.


sjboeing (approved):
Looks good!

matt-frey:
This issue / PR is connected to what we experienced in #540.

PR #596, opened 2025-01-24
Fix error in get_local_bounds

matt-frey:
The variable `last` is passed with `intent(out)`. We cannot use the variable before it is assigned. Also, this lines has no effect because a value is assigned to the variable at the end of the routine.


sjboeing (approved):
Sounds good!

PR #597, opened 2025-03-11
Fix label of planetary vorticity

matt-frey:
There is a label mismatch for the panetary vorticity input, which results in ignoring the flag at restarts.


sjboeing (approved):
Yes, this seems more consistent.

PR #602, opened 2025-03-27
Fix issue when MPI ranks agree but MPI decomposition is different

matt-frey:
When we write parcel files and then restart a simulation, we must ensure the MPI decomposition is identical. Otherwise, this leads to an incorrect reading of parcel data.


sjboeing (approved):
Thanks, Matthias! Looks like a good way to check this in general.

